PR #54, opened 2022-06-22
Add HttpClient Extension

C-Christiansen:


libraries/dotnet/EnergyOriginAuthorization/EnergyOriginAuthorization.Tests/HttpClientExtensionTests.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        var client = new HttpClient();
+        var context = new AuthorizationContext("subject", "actor", "token");
+
+        client.AddAuthorizationToken(context);
+
+        Assert.NotNull(client);
+        Assert.NotEmpty(client.DefaultRequestHeaders);
+        Assert.Equal(context.Token, client.DefaultRequestHeaders.Authorization?.Parameter ?? "");
+        Assert.Equal("Bearer", client.DefaultRequestHeaders.Authorization?.Scheme ?? "");
```
Exobitt, line 20: Why are these allowed nullable types?
As in token, shouldn't there always be some kind of token, or else get denied in-front?

Exobitt (changes requested):

libraries/dotnet/EnergyOriginAuthorization/EnergyOriginAuthorization.Tests/HttpClientExtensionTests.cs
CodeReaper: @Exobitt The nullable type is from `HttpClient` in `System.Net.Http`.

Exobitt (approved):

PR #103, opened 2022-07-25
Preview/auth domain

Exobitt:


domains/auth/Auth.API/API/Models/Oidc/InternalToken.cs
```diff
@@ -0,0 +1,18 @@
...
+    public List<string> Scope { get; }
+
+    public InternalToken(DateTime issued, DateTime expires, string actor, string subject, List<string> scope)
+    {
+        Issued = issued;
+        Expires = expires;
+        Actor = actor;
+        Subject = subject;
+        Scope = scope;
+    }
```
MartinSchmidt, line 17:
```suggestion
public record InternalToken
{
    public DateTime Issued { get; init; }
    public DateTime Expires { get; init; }
    public string Actor { get; init; }
    public string Subject { get; init; }
    public List<string> Scope { get; init; }
```

domains/auth/chart/values.yaml
```diff
@@ -0,0 +1,14 @@
...
+    tag: 0.0.1
+  env:
+    INTERNALTOKENSECRET
+    TOKENEXPIRYTIME
+    SCOPE
+    AMRVALUES
+    OIDCURL
+    OIDCCLIENTID
+    OIDCCLIENTSECRET
+    BASEURL
```
MartinSchmidt, line 14:
```suggestion
```

these should be removed. setting ENV this way, does not enable us to override them using configmaps from eo-base-environment.

domains/auth/Auth.API/Tests/Resources/AddEnvironmentVariables.cs
```diff
@@ -0,0 +1,18 @@
+using System;
```
MartinSchmidt, line 1: When testing with environment variables have a look at how to ensure they are cleaned up at:  https://github.com/Energinet-DataHub/eo-data-sync/blob/main/services/DataSync.API/Tests/Helper/EnvironmentBase.cs

and here for seriel execution to ensure that unittests with different variables don't accidentially share environment variables.   [  [Collection("environment-tests")]](https://github.com/Energinet-DataHub/eo-data-sync/blob/c8393d97708615581a97a7138a6355b22a79e1d7/services/DataSync.API/Tests/Helper/ConfigurationTests.cs#L8)

domains/auth/Auth.API/API/Services/TokenService.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    public QueryBuilder CreateAuthorizationRedirectUrl(string responseType, string feUrl, AuthState state, string lang)
+    {
+        var query = new QueryBuilder();
+
+        query.Add("response_type", responseType);
+        query.Add("client_id", Configuration.GetOidcClientId());
+        query.Add("redirect_uri", $"{feUrl}/api/auth/oidc/login/callback");
+        query.Add("scope", Configuration.GetScopes());
+        query.Add("state", EncodeJwtToken(state));
```
MartinSchmidt, line 52: We should reconsider this, the state should preferably be encrypted and not just signed in a JWT.

domains/auth/Auth.API/API/Services/SignaturGruppen/ISignaturGruppen.cs
```diff
@@ -0,0 +1,7 @@
+using API.Models;
+
+namespace API.Services;
+public interface ISignaturGruppen
+{
+    LoginResponse CreateRedirecthUrl(string feUrl, string returnUrl);
```
MartinSchmidt, line 6: I would change this to GetNext(AuthState state)

and then create a new empty state based on these parameters in the login controller to start the flow, then we would already have the basis for the orchestrator done.

MartinSchmidt (changes requested):
- Rename SignaturGruppen to orchestrator

domains/auth/Auth.API/API/Services/CryptographyService.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Text;
+
+namespace API.Services;
+public class CryptographyService : ICryptographyService
+{
+    string key = Configuration.GetSecretKey();
+
+    public string EncryptState(AuthState state)
+    {
+        byte[] iv = new byte[16];
```
MartinSchmidt, line 13: IV should not be an empty array, when a new Aes is created, and random IV is automatically generated, often one would take the 16 bytes from IV and use as the first 16 bytes of the data package.

This could be done on line 20 by writing iv to the memory stream before the cryptostream.

domains/auth/Auth.API/API/Services/CryptographyService.cs
```diff
@@ -0,0 +1,56 @@
...
+            {
+                using (CryptoStream cryptoStream = new CryptoStream(memoryStream, encryptor, CryptoStreamMode.Write))
+                {
+                    using (StreamWriter streamWriter = new StreamWriter(cryptoStream))
+                    {
+                        streamWriter.Write(state.ToString());
+                    }
+                    byte[] data = memoryStream.ToArray();
+                    return Convert.ToBase64String(data);
+                }
```
MartinSchmidt, line 29:
```suggestion
                }
                byte[] data = memoryStream.ToArray();
                return Convert.ToBase64String(data);
```
i would suggest moving this outside the cryptostream.

domains/auth/Auth.API/API/Services/CryptographyService.cs
```diff
@@ -0,0 +1,56 @@
...
+            {
+                using (CryptoStream cryptoStream = new CryptoStream(memoryStream, decryptor, CryptoStreamMode.Read))
+                {
+                    using (StreamReader streamReader = new StreamReader(cryptoStream))
+                    {
+                        return streamReader.ReadToEnd();
+                    }
+                }
+            }
+        }
```
MartinSchmidt, line 54: Similar, the iv should be read from the first 16 bytes

```suggestion
        byte[] buffer = Convert.FromBase64String(encryptedState);

        using (MemoryStream memoryStream = new MemoryStream(buffer))
        {
            var iv = memoryStream.read(16);
            using (Aes aes = Aes.Create())
            {
                aes.Key = Encoding.UTF8.GetBytes(key);
                aes.IV = iv;
                ICryptoTransform decryptor = aes.CreateDecryptor(aes.Key, aes.IV);
            
                using (CryptoStream cryptoStream = new CryptoStream(memoryStream, decryptor, CryptoStreamMode.Read))
                {
                    using (StreamReader streamReader = new StreamReader(cryptoStream))
                    {
                        return streamReader.ReadToEnd();
                    }
                }
            }
        }
    }
```

MartinSchmidt (changes requested):
- I do not see any tests for the cryptography service, i think it would be a good idea to write seperate unit tests to document it working as expected.

domains/auth/Auth.API/API/Services/CryptographyService.cs
```diff
@@ -0,0 +1,56 @@
...
+    public string EncryptState(AuthState state)
+    {
+        byte[] iv = new byte[16];
+        using (Aes aes = Aes.Create())
+        {
+            aes.Key = Encoding.UTF8.GetBytes(key);
+            aes.IV = iv;
+            ICryptoTransform encryptor = aes.CreateEncryptor(aes.Key, aes.IV);
+            using (MemoryStream memoryStream = new MemoryStream())
+            {
```
MartinSchmidt, line 20:
```suggestion
        using (Aes aes = Aes.Create())
        {
            aes.Key = Encoding.UTF8.GetBytes(key);
             = iv;
            ICryptoTransform encryptor = aes.CreateEncryptor(aes.Key, aes.IV);
            using (MemoryStream memoryStream = new MemoryStream())
            {
                memoryStream.write(aes.IV);
```

domains/auth/Auth.API/API/Services/CryptographyService.cs
```diff
@@ -10,39 +10,44 @@ public class CryptographyService : ICryptographyService
...
                     {
                         streamWriter.Write(state.ToString());
                     }
-                    byte[] data = memoryStream.ToArray();
-                    return Convert.ToBase64String(data);
+                    data = memoryStream.ToArray();
+                    
                 }
             }
+            return Convert.ToBase64String(data);
```
MartinSchmidt, line 33: The reason to wait until after closing the using for cryptostream is to ensure it has flushed its write to the memory streams.

```suggestion
                 }
                 return Convert.ToBase64String(memoryStream.ToArray());
             }
```

domains/auth/Auth.API/API/Services/CryptographyService.cs
```diff
@@ -10,39 +10,44 @@ public class CryptographyService : ICryptographyService
 
     public string EncryptState(AuthState state)
     {
-        byte[] iv = new byte[16];
+        byte[] data;
+        //byte[] iv = new byte[16];
```
MartinSchmidt, line 14:
```suggestion
```

domains/auth/Auth.API/API/Services/CryptographyService.cs
```diff
@@ -10,39 +10,44 @@ public class CryptographyService : ICryptographyService
...
                 }
             }
+            return Convert.ToBase64String(data);
         }
     }
 
     public string decryptState(string encryptedState)
     {
-        byte[] iv = new byte[16];
+        //byte[] iv = new byte[16];
```
MartinSchmidt, line 39: not used
```suggestion
```

domains/auth/Auth.API/API/Services/CryptographyService.cs
```diff
@@ -10,39 +10,44 @@ public class CryptographyService : ICryptographyService
 
     public string EncryptState(AuthState state)
```
MartinSchmidt, line 11: Either we should move to a generic Encrypt function that encryptes a string and returns a string, similar to decrypt, or decrypt should specifically return an AuthState.

I would lean towards generic Encrypt(string data)

domains/auth/Auth.API/Tests/TestCryptographics.cs
```diff
@@ -0,0 +1,54 @@
...
+namespace Tests;
+
+[UnitTest]
+public sealed class TestCryptographics
+{
+    private readonly Mock<ICryptographyService> _cryptographyService;
+
+    public TestCryptographics()
+    {
+        _cryptographyService = new Mock<ICryptographyService>();
```
MartinSchmidt, line 18: Do you create a mock?

Why? are we not specifically interested in testing your implementation,

I would write a test like:

```c#
var cryptoserver = new CryptographyService()

var data = "some random text"
var encrypted = cryptoserver.Encrypt(data);
var decrypted = cryptoserver.Decrypt(encrypted);

Assert.NotEqual(data, encrypted);
Assert.Equal(data, decrypted);
```

MartinSchmidt (changes requested):
I hope the things make sense...

domains/auth/Auth.API/API/Services/CryptographyService.cs
```diff
@@ -0,0 +1,58 @@
...
+            ICryptoTransform encryptor = aes.CreateEncryptor(aes.Key, aes.IV);
+            using (MemoryStream memoryStream = new MemoryStream())
+            {
+                memoryStream.Write(aes.IV);
+
+                using (CryptoStream cryptoStream = new CryptoStream(memoryStream, encryptor, CryptoStreamMode.Write))
+                {
+                    using (StreamWriter streamWriter = new StreamWriter(cryptoStream))
+                    {
+                        streamWriter.Write(state.ToString());
```
MartinSchmidt, line 25: .TOSTRING.

domains/auth/Auth.API/API/Services/CryptographyService.cs
```diff
@@ -0,0 +1,58 @@
...
+using API.Helpers;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace API.Services;
+public class CryptographyService : ICryptographyService
+{
+    string key = Configuration.GetSecretKey();
+
+    public string EncryptState(string state)
```
MartinSchmidt, line 11: EncryptState -> Encrypt

domains/auth/Auth.API/API/Services/CryptographyService.cs
```diff
@@ -0,0 +1,58 @@
...
+                    {
+                        streamWriter.Write(state.ToString());
+                    }
+                }
+                return Convert.ToBase64String(memoryStream.ToArray());
+            }
+        }
+    }
+
+    public string decryptState(string encryptedState)
```
MartinSchmidt, line 33: decryptState -> Decrypt

domains/auth/Auth.API/API/Services/CryptographyService.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    public string EncryptState(string state)
+    {
+        using (Aes aes = Aes.Create())
+        {
+            aes.Key = Encoding.UTF8.GetBytes(key);
+            ICryptoTransform encryptor = aes.CreateEncryptor(aes.Key, aes.IV);
+            using (MemoryStream memoryStream = new MemoryStream())
+            {
+                memoryStream.Write(aes.IV);
```
MartinSchmidt, line 19: remove this!

MartinSchmidt (changes requested):

MartinSchmidt (approved):

PR #106, opened 2022-08-10
Add in-memory event store

CodeReaper:


Exobitt (approved):

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MemoryEventStore.cs
```diff
@@ -0,0 +1,133 @@
...
+                    consumer.OnMessage(this, message);
+                });
+            }
+
+            _actions.Enqueue(() =>
+            {
+                OnMessage += consumer.OnMessage;
+            });
+
+            await Drain().ConfigureAwait(false);
```
duizer, line 35: Is this call to Drain() needed as it also on line 38?

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MessageEventArgs.cs
```diff
@@ -0,0 +1,16 @@
+
+namespace EnergyOriginEventStore.EventStore.Memory;
+
+public class MessageEventArgs : EventArgs
```
duizer, line 4: Believe this class should be internal

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MemoryEventStore.cs
```diff
@@ -0,0 +1,133 @@
...
+                OnMessage?.Invoke(this, item);
+            });
+        }
+
+        await Drain().ConfigureAwait(false);
+    }
+
+    public IEventConsumerBuilder GetBuilder(string topicPrefix) => new MemoryEventConsumerBuilder(this, topicPrefix);
+
+    public event Action? DisposeEvent;
```
duizer, line 63: There is (in this PR) no usage of DisposeEvent. Is it actually needed?
Also, since it is an "Action" and not an "event", so the name should be something like "DisposeAction"

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MemoryEventStore.cs
```diff
@@ -0,0 +1,133 @@
...
+        {
+            await Drain();
+        }
+    }
+
+    #endregion
+
+    #region Pointers
+
+    private string Pointer(InternalEvent model) => $"{model.Issued}-{model.IssuedFraction}";
```
duizer, line 95:
```suggestion
    private string ToPointer(InternalEvent model) => $"{model.Issued}-{model.IssuedFraction}";
```

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MemoryEventStore.cs
CodeReaper: Yes, L38 starts the action that enqueues the real work, the L35 starts the real work actions that was just enqueued.

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MemoryEventStore.cs
CodeReaper: It is required by the EventStore interface:
```
public interface IEventStore : IDisposable
```
Should it be required?

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MemoryEventStore.cs
```diff
@@ -0,0 +1,133 @@
...
+
+    internal bool ShouldInclude(string? continuationPointer, string messagePointer)
+    {
+        if (continuationPointer == null) return true;
+
+        var continuationValues = PointerValues(continuationPointer);
+        var messageValues = PointerValues(messagePointer);
+
+        return messageValues.Item1 > continuationValues.Item1 || (messageValues.Item1 == continuationValues.Item1 && messageValues.Item2 > continuationValues.Item2);
+    }
```
duizer, line 122: I believe using tuple types makes this a bit easier to read:

```suggestion
    private static (long issued, long fraction) PointerValues(string pointer)
    {
        long issued;
        long fraction;
        try
        {
            var parts = pointer.Split('-');
            issued = long.Parse(parts[0]);
            fraction = long.Parse(parts[1]);
        }
        catch (Exception)
        {
            throw new InvalidDataException($"Pointer '{pointer}' not a valid format");
        }
        return (issued, fraction);
    }

    internal bool ShouldInclude(string? continuationPointer, string messagePointer)
    {
        if (continuationPointer == null) return true;

        var continuationValues = PointerValues(continuationPointer);
        var messageValues = PointerValues(messagePointer);

        return messageValues.issued > continuationValues.issued || (messageValues.issued == continuationValues.issued && messageValues.fraction > continuationValues.fraction);
    }
```

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MemoryEventStore.cs
duizer: As far as I can see it is not part of the interface:

```
public interface IEventStore : IDisposable
{
    Task Produce(EventModel model, params string[] topics);

    IEventConsumerBuilder GetBuilder(string topicPrefix);
}
```

```
public interface IDisposable
{
    void Dispose();
}
```

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MemoryEventStore.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        if (!_actions.IsEmpty)
+        {
+            await Drain();
+        }
+    }
+
+    #endregion
+
+    #region Pointers
```
duizer, line 93: This is an idea and let me know what you think: 
We create a new class called "MemoryPointer", which has the functionality in the "Pointers" region in MemoryEventStore. This will remove some 30 lines from MemoryEventStore

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MemoryEventStore.cs
```diff
@@ -0,0 +1,133 @@
...
+using EnergyOriginEventStore.EventStore.Serialization;
+using Newtonsoft.Json;
+
+namespace EnergyOriginEventStore.EventStore.Memory;
+
+public class MemoryEventStore : IEventStore
+{
+    private List<MessageEventArgs> _messages = new();
+    private ConcurrentQueue<Action> _actions = new();
+    [ThreadStatic] private bool isDraining = false;
```
duizer, line 11: See https://www.jetbrains.com/help/resharper/ThreadStaticAtInstanceField.html for more info

```suggestion
    [ThreadStatic] private static bool isDraining = false;
```

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore.Tests/MemoryEventStoreTests.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        await Task.Delay(TimeSpan.FromMilliseconds(50));
+
+        Assert.Single(received);
+    }
+
+    [Fact]
+    public async Task EventStore_MultipleListers_success()
+    {
+        var semaphore = new SemaphoreSlim(0);
```
duizer, line 114: This line can be deleted as `semaphore` is not used.

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MemoryEventStore.cs
duizer: Since it is used to lock a critical section with async/await keywords, then this should be a `new SemaphoreSlim(1,1)`

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MemoryEventConsumer.cs
```diff
@@ -0,0 +1,40 @@
...
+
+    public MemoryEventConsumer(IUnpacker unpacker, Dictionary<Type, IEnumerable<Action<Event<EventModel>>>> handlers, MemoryEventStore store, string topicPrefix, MemoryPointer? pointer)
+    {
+        _store = store;
+        _unpacker = unpacker;
+        _handlers = handlers;
+        _topicPrefix = topicPrefix;
+        _pointer = pointer;
+
+        Task.Run(async () => await _store.Attach(this).ConfigureAwait(false));
```
duizer, line 22: Generally the constructor should be fast and not perform actual work. Furthermore, since it is not possible to have an async constructor, there is a need to do Task.Run().
But if we create a static method called e.g. `Create` we can let that method be async and then call `.Attach()` on MemoryEventStore.

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore.Tests/MemoryEventStoreTests.cs
```diff
@@ -0,0 +1,145 @@
...
+        IEventStore eventStore = new MemoryEventStore();
+
+        var message = new Said("Anton Actor", "I like to act!");
+        await eventStore.Produce(message, "Gossip", "Tabloid");
+
+        Said? receivedValue = null;
+        eventStore
+            .GetBuilder("Gossip")
+            .AddHandler<Said>((value) =>
```
duizer, line 22:
```suggestion
            .AddHandler<Said>(value =>
```

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore.Tests/MemoryEventStoreTests.cs
```diff
@@ -0,0 +1,145 @@
...
+        await Task.Delay(TimeSpan.FromMilliseconds(50));
+
+        Assert.Single(received);
+        Assert.Equal(message3, received.Single().Statement);
+    }
+
+    [Fact]
+    public async Task EventStore_FiltersTopics_success()
+    {
+        var semaphore = new SemaphoreSlim(0);
```
duizer, line 94: Not used. Can be delted

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore.Tests/MemoryEventStoreTests.cs
```diff
@@ -0,0 +1,145 @@
...
+        IEventStore eventStore = new MemoryEventStore();
+
+        var message = new Said("Samuel Salesman", "We have been trying to reach you about your cars extended warrenty!");
```
duizer, line 97:
```suggestion
        var message = new Said("Samuel Salesman", "We have been trying to reach you about your cars extended warranty!");
```

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore.Tests/MemoryEventStoreTests.cs
```diff
@@ -0,0 +1,145 @@
...
+            .AddHandler<Said>(value => received.Add(value.EventModel))
+            .Build();
+
+        await Task.Delay(TimeSpan.FromMilliseconds(50));
+
+        Assert.Single(received);
+    }
+
+    [Fact]
+    public async Task EventStore_MultipleListers_success()
```
duizer, line 112:
```suggestion
    public async Task EventStore_MultipleListeners_success()
```

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MessageEventArgs.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    public readonly string Message;
+    public readonly string Topic;
+    public readonly MemoryPointer Pointer;
+
+    internal MessageEventArgs(string message, string topic, MemoryPointer pointer)
+    {
+        Message = message;
+        Topic = topic;
+        Pointer = pointer;
```
duizer, line 14:
```suggestion
        Message = message ?? throw new ArgumentNullException(nameof(message));
        Topic = topic ?? throw new ArgumentNullException(nameof(topic));
        Pointer = pointer ?? throw new ArgumentNullException(nameof(pointer));
```

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MemoryEventStore.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+
+    #endregion
+
+    #region Events
+
+    internal event MessageEventHandler? OnMessage;
+
+    internal delegate void MessageEventHandler(object sender, MessageEventArgs e);
```
duizer, line 94: This could be changed to `internal event EventHandler<MessageEventArgs> OnMessage;`
See https://docs.microsoft.com/en-us/dotnet/standard/events/how-to-raise-and-consume-events#example-2 for example

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MemoryEventStore.cs
```diff
@@ -0,0 +1,97 @@
+using System.Collections.Concurrent;
+using EnergyOriginEventStore.EventStore.Serialization;
+using Newtonsoft.Json;
+
+namespace EnergyOriginEventStore.EventStore.Memory;
+
+public class MemoryEventStore : IEventStore
+{
+    private List<MessageEventArgs> _messages = new();
+    private ConcurrentQueue<Action> _actions = new();
```
duizer, line 10:
```suggestion
    private readonly List<MessageEventArgs> _messages = new();
    private readonly ConcurrentQueue<Action> _actions = new();
```

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MessageEventArgs.cs
duizer: This could avoided by treating this as an error and not a warning. Add the following to editorconfig

```
# CS8625: Cannot convert null literal to non-nullable reference type.
dotnet_diagnostic.CS8625.severity = error
```

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore.Tests/MemoryEventStoreTests.cs
```diff
@@ -0,0 +1,145 @@
+using EnergyOriginEventStore.EventStore;
+using EnergyOriginEventStore.EventStore.Memory;
+using EnergyOriginEventStore.Tests.Topics;
+using Xunit;
+
+namespace EnergyOriginEventStore.Tests;
+
+public class MemoryEventStoreTests
```
duizer, line 8: We should add a test that performs the following steps:

1. Creates consumer and ensures that it is drained (using await Thread.Delay)
2. Produces a new message
3. Asserts that the consumer created in step 1 receives the message produced in step 2.

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MemoryEventConsumer.cs
CodeReaper: We are going to leave this as is and see what changes we potentially need in the interface. We may want to have the `build()` method be async.

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MessageEventArgs.cs
CodeReaper: Issue created

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MemoryEventStore.cs
```diff
@@ -0,0 +1,91 @@
...
+using EnergyOriginEventStore.EventStore.Serialization;
+using Newtonsoft.Json;
+
+namespace EnergyOriginEventStore.EventStore.Memory;
+
+public class MemoryEventStore : IEventStore
+{
+    private readonly List<MessageEventArgs> _messages = new();
+    private readonly ConcurrentQueue<Action> _actions = new();
+    private SemaphoreSlim _drainSemaphore = new SemaphoreSlim(1, 1);
```
duizer, line 11:
```suggestion
    private static SemaphoreSlim _drainSemaphore = new SemaphoreSlim(1, 1);
```

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore/EventStore/Memory/MemoryEventStore.cs
```diff
@@ -0,0 +1,91 @@
...
+    public void Dispose()
+    {
+        _actions.Clear();
+    }
+
+    private async Task Drain()
+    {
+        if (_drainSemaphore.CurrentCount == 0) return;
+
+        _drainSemaphore.Wait();
```
duizer, line 74: With this we achieve that we will not make unnecessary calls to Drain. And I think that is fine. 
But I want to mention, that there is an async overload `WaitAsync()` which could be used instead. I also believe that if change line 72 - 74 to a single line of `await _drainSemapore.WaitAsync();`, then we do not have to have the check at line 84 if `_actions` is empty.

libraries/dotnet/EnergyOriginEventStore/EnergyOriginEventStore.Tests/FlatFileEventStoreTests.cs
```diff
@@ -95,6 +91,64 @@ public async Task EventStore_ResumeFromPointer_success()
...
+
+        var message = new Said("Exavier Exception", "I have eruptive things to say!");
+        await eventStore.Produce(message, "Unstable");
+
+        eventStore
+            .GetBuilder("Unstable")
+            .AddHandler<Said>(_ => throw new NotImplementedException("Oh Exavier did it again..."))
+            .Build();
+
+        await Task.Delay(TimeSpan.FromMilliseconds(50));
```
duizer, line 108: Same thing here as my comment in MemoryEventStoreTest

duizer (approved):

PR #108, opened 2022-08-16
WIP: Add sonar cloud scanning workflow

Sondergaard:
Copy / paste of default build.yml from Sonar Cloud


CodeReaper:
There seems to an action you can use for sonarcloud scanning - see https://github.com/SonarSource/sonarcloud-github-action

However the workflow probably needs to adjusted to scan all domains and libraries regularly or only the changed ones if it is supposed to run on PRs.

PR #117, opened 2022-08-19
Auth logout

C-Christiansen:


domains/auth/Auth.API/API/Controllers/AuthController.cs
```diff
@@ -2,20 +2,27 @@
...
 namespace API.Controllers;
 
 
 [ApiController]
-public class AuthController
+public class AuthController : ControllerBase
 {
     readonly ILogger<AuthController> logger;
     readonly IOidcProviders oidcProviders;
+    readonly ICookieService cookieService;
```
Exobitt, line 16: Rename these to have an underscore(_) in front of their name. 
This is due to they're only used in this class

domains/auth/Auth.API/API/Controllers/AuthController.cs
```diff
@@ -33,4 +40,20 @@ public AuthController(ILogger<AuthController> logger, IOidcProviders oidcProvide
         return oidcProviders.CreateAuthorizationUri(state);
     }
 
+    [HttpPost]
+    [Route("/auth/logout")]
+    public ActionResult<LogoutResponse> Logout()
+    {
+
```
Exobitt, line 47: remove spacing

domains/auth/Auth.API/Tests/TestLogout.cs
```diff
@@ -0,0 +1,61 @@
...
+using Xunit;
+using Xunit.Categories;
+using Moq;
+using API.Controllers;
+using Microsoft.Extensions.Logging;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Http;
+using System.Linq;
+
+namespace Tests;
```
ckr123, line 13: Should we arrange this in a Test.Service dir so we can get some structure in our test projects.?

domains/auth/Auth.API/Tests/TestLogout.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        var cookieService = new CookieService(authOptionsMock.Object);
+
+        var testToken = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3N1ZWQiOiIyMDIyLTA1LTE4VDEzOjEzOjQxLjY2MTI2NSswMDowMCIsImV4cGlyZXMiOiIyMDIyLTA1LTE5VDEzOjEzOjQxLjY2MDk4NyswMDowMCIsImFjdG9yIjoiYWN0b3IiLCJzdWJqZWN0Ijoic3ViamVjdCIsInNjb3BlIjpbInNjb3BlMSIsInNjb3BlMiJdfQ.W1C1xKiEYPDeuo1OfRFpm6L3j7YGQJTGmegIgLu2UIQ";
+        var opaqueToken = "TestOpaqueToken";
+
+        //Act
+        var cookieOptions = cookieService.CreateCookieOptions(authOptionsMock.Object.Value.CookieCreateExpires);
+        
+        var AuthController = new AuthController(logger.Object, odieProvider.Object, cookieService, authOptionsMock.Object) { ControllerContext = new ControllerContext() { HttpContext = new DefaultHttpContext() }};
```
ckr123, line 46:
```suggestion
        var AuthController = new AuthController(
        logger.Object, 
        odieProvider.Object, 
        cookieService, 
        authOptionsMock.Object) {
         ControllerContext = new ControllerContext() 
         { HttpContext = new DefaultHttpContext() 
         }};
```
Would it make it more readable to split it into multiple lines? I think it would make sense if we define a hard line break in our code style

domains/auth/Auth.API/Tests/TestLogout.cs
C-Christiansen: Should'nt be Test.Controller then?

domains/auth/Auth.API/API/Controllers/AuthController.cs
```diff
@@ -30,7 +34,22 @@ public AuthController(ILogger<AuthController> logger, IOidcProviders oidcProvide
...
+    [HttpPost]
+    [Route("/auth/logout")]
+    public ActionResult<LogoutResponse> Logout()
+    {
+        var token = HttpContext.Request.Headers[_authOptions.CookieName].FirstOrDefault()?.Split(" ").Last();
+
+        if (token != null)
+        {
+            //delete token from database
+            //oidcProviders.Logout(token);
```
duizer, line 49: Why is this commented out?

Exobitt (approved):

domains/auth/Auth.API/API/Models/Logout.cs
```diff
@@ -0,0 +1,8 @@
+namespace API.Models;
+
+public record LogoutResponse
+{
+    public bool success { get; init; }
+}
+
+
```
duizer, line 8: This empty line can be deleted so there is only 1

domains/auth/chart/values.yaml
```diff
@@ -2,5 +2,4 @@ api:
   replicaCount: 3
   image:
     name: ghcr.io/energinet-datahub/eo-auth
-    tag: 0.0.3
-
```
duizer, line 6: There should be an empty newline here

domains/auth/Auth.API/Tests/TestAuthController.cs
```diff
@@ -0,0 +1,66 @@
...
+        var AuthController = new AuthController(logger.Object, oidcProvider.Object, authOptionsMock.Object)
+        {
+            ControllerContext = new ControllerContext()
+            {
+                HttpContext = new DefaultHttpContext()
+            }
+        };
+
+        AuthController.HttpContext.Response.Cookies.Append("Authorization", $"{opaqueToken}", cookieOptions);
+        AuthController.HttpContext.Request.Headers.Add("Authorization", "Bearer " + testToken);
```
duizer, line 59: To make the test easier to read you can inline `testToken` and then also use a simpler example than an actual JWT token. I think "foo" is fine

domains/auth/Auth.API/Tests/TestAuthController.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        AuthController.Logout();
+
+        //Assert
+        Assert.Equal(expectedCookie, AuthController.HttpContext.Response.Headers.Values.Single());
```
duizer, line 64: I missed that `expectedCookie` is an expired cookie. Either inline it here or change name of `expectedCookie` to `expectedExpiredCookie` or `expiredCookie`

domains/auth/Auth.API/API/Controllers/AuthController.cs
C-Christiansen: These are outcommented due to their functions havent been implemted yet.

duizer (approved):

PR #120, opened 2022-08-23
my proposal for a structure in the service

ckr123:
After a small talk with @CodeReaper and @C-Christiansen  i made this proposal on how i suggest we structure the auth service.
Hopefully this PR can be used to agree on a common structure in our services and avoid future problems when starting on a new service with multiple developers :)

I would suggest we look into some patterns eg. my favourite [Hexagonial Architecture](https://alistair.cockburn.us/hexagonal-architecture/) as i think right know we need a common base our controllers can be dependent on and not different repo/services/providers :)


domains/auth/Auth.API/Tests/Controllers/TestAuthController.cs
```diff
@@ -83,39 +81,39 @@ public void ReturnBadRequestWhenNoToken()
             .Setup(x => x.Decrypt(authStateAsString))
             .Returns(authStateAsString);
 
-        var response = _authController.Invalidate(authStateAsString);
+        var response = _invalidateController.Invalidate(authStateAsString);
 
         Assert.IsType<BadRequestObjectResult>(response);
         var result = response as BadRequestObjectResult;
         Assert.Equal(nameof(AuthState.IdToken) + " must not be null", result?.Value);
     }
```
CodeReaper, line 90: Out commented, because?...

domains/auth/Auth.API/API/Controllers/InvalidateController.cs
```diff
@@ -0,0 +1,54 @@
...
+        IValidator<AuthState> validator)
+    {
+        _oidcService = oidcService;
+        _validator = validator;
+        _cryptography = cryptography;
+    }
+
+    [HttpPost]
+    [Route("/invalidate")]
+    public IActionResult Invalidate([FromQuery] string state)
```
duizer, line 28: What about using [FromServices] for injection of dependencies: https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/dependency-injection?view=aspnetcore-6.0#action-injection-with-fromservices

ckr123: I will rather have this in another PR. 
Right now the controller max has 3 dependencies, but i  think it will be a great solution if we have to handle more dependencies so we can overvoid an overloaded constructor.

Will that be alright for you?

duizer: That's fine

C-Christiansen (approved):

domains/auth/Auth.API/API/Services/OidcProviders/SignaturGruppen.cs
```diff
@@ -59,4 +61,21 @@ public async Task Logout(string token)
...
+
+        var query = new QueryBuilder
+        {
+            { "response_type", responseType },
+            { "client_id", authOptions.OidcClientId },
+            { "redirect_uri", $"{state.FeUrl}/api/auth/oidc/login/callback" },
+            { "scope", authOptions.Scope },
+            { "state", cryptography.Encrypt(serilizedJson) },
+            { "language", lang }
+        };
```
CodeReaper, line 77:
```suggestion
        var json = JsonSerializer.Serialize(state);

        var query = new QueryBuilder
        {
            { "response_type", responseType },
            { "client_id", authOptions.OidcClientId },
            { "redirect_uri", $"{state.FeUrl}/api/auth/oidc/login/callback" },
            { "scope", authOptions.Scope },
            { "state", cryptography.Encrypt(json) },
            { "language", lang }
        };
```

domains/auth/Auth.API/Tests/Services/OidcProviders/TestSignaturGruppen.cs
```diff
@@ -66,4 +72,28 @@ public async void CannotLogoutFromSignaturGruppenSoWeLogAMessage()
                 It.IsAny<Func<It.IsAnyType, Exception, string>>()!),
             Times.Once);
     }
+
+    [Fact]
+    public void RedirectSuccess()
+    {
+        //Arrange
```
CodeReaper, line 79:
```suggestion
```

CodeReaper (approved):

PR #121, opened 2022-08-24
Add dotnet syntax help

CodeReaper:


domains/auth/Auth.API/API/Controllers/AuthController.cs
```diff
@@ -11,17 +11,17 @@ namespace API.Controllers;
 [ApiController]
 public class AuthController : ControllerBase
 {
-    readonly ILogger<AuthController> _logger;
-    readonly IOidcProviders _oidcProviders;
-    readonly ITokenStorage _tokenStorage;
-    private readonly AuthOptions _authOptions;
+    readonly ILogger<AuthController> logger;
```
pleuph, line 14: These should be unified so either none or all have the private modifier.

domains/auth/Auth.API/API/Controllers/AuthController.cs
```diff
@@ -30,29 +30,29 @@ public AuthController(ILogger<AuthController> logger, IOidcProviders oidcProvide
         [Required] string feUrl,
         [Required] string returnUrl)
     {
-        AuthState state = new AuthState()
+        var state = new AuthState()
```
pleuph, line 33: Parentheses can be omitted when initializing in this way. Can this be automated?

libraries/dotnet/EnergyOriginAuthorization/EnergyOriginAuthorization/EnergyOriginAuthorization/AuthorizeAttribute.cs
```diff
@@ -7,29 +7,29 @@ namespace EnergyOriginAuthorization
...
         public AuthorizeAttribute()
         {
-            _requiredScopes = new List<string>();
+            requiredScopes = new List<string>();
         }
 
         public AuthorizeAttribute(string requiredScope)
         {
-            _requiredScopes = new List<string>() { requiredScope };
+            requiredScopes = new List<string>() { requiredScope };
```
pleuph, line 20: Parentheses can be omitted.

domains/auth/Auth.API/API/Services/OidcService.cs
```diff
@@ -1,22 +1,22 @@
...
-using System.Text.Json;
 
 namespace API.Services;
 
 public class OidcService : IOidcService
 {
-    readonly ILogger _logger;
-    readonly ICryptographyService _cryptography;
-    private readonly AuthOptions _authOptions;
+    readonly ILogger logger;
```
pleuph, line 11: Unify modifiers

pleuph (changes requested):
A few minor changes to be done, and one to investigate.

domains/auth/Auth.API/API/Controllers/AuthController.cs
CodeReaper: There appears to be only [two rules](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0047-ide0048) for parentheses and them both seem irrelevant in regards to this. If we should find more rules, let us add them in another PR.

domains/measurements/Measurements.API/API/Helpers/Configuration.cs
```diff
@@ -2,11 +2,11 @@ namespace API.Helpers
 {
     public static class Configuration
     {
-        const string DataSyncEndpoint = "DATASYNCENDPOINT";
+        const string dataSyncEndpoint = "DATASYNCENDPOINT";
```
robertrosborg, line 5: Shouldn't this be PascalCase?

CodeReaper: @robertrosborg - The rules we have setup up says the public and protected should be PascalCase and everything else is camelCase (rules: `dotnet_naming*`). That string is (implicitly) private. We did discuss enforcing that implicit modifiers always be present (rule: `dotnet_style_require_accessibility_modifiers`), but opt (so far) for having it set to the default.

.editorconfig
```diff
@@ -6,8 +6,8 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
-[*.cs]
-indent_size = 4
+[*.yaml]
+indent_size = 2
 indent_style = space
```
ckr123, line 11: Would it make sense to use `indent_style = space` as a default setting `[*]`?

.editorconfig
```diff
@@ -18,6 +18,129 @@ indent_style = space
...
 
-[*.yaml]
-indent_size = 2
+[*.cs]
+indent_size = 4
 indent_style = space
+
+# New line preferences
+csharp_new_line_before_open_brace = all
```
ckr123, line 26: Is there a place where i can see an example of these styles?

.editorconfig
CodeReaper: @ckr123 The first tooling we used would apply the settings to everything including binary files, so I would suggest only configuring what we want to configure.

.editorconfig
CodeReaper: Yes, here is [a link](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/csharp-formatting-options).

ckr123 (approved):

PR #131, opened 2022-08-31
Refactor to features

MartinSchmidt:


domains/emissions/Emissions.API/API/Extensions/IServiceCollectionExtension.cs
```diff
@@ -0,0 +1,19 @@
...
+using API.Features.EnergySources;
+using API.Services;
+using API.Shared.DataSync;
+using API.Shared.EnergiDataService;
+
+namespace API.Extensions;
+
+public static class IServiceCollectionExtension
+{
+    public static void AddCustomServices(this IServiceCollection services)
```
TopSwagCode, line 12: Would perhaps have other name than Custom?

PR #141, opened 2022-10-19
add new domain

ckr123:
Add wepapi domain with "hello world" example


duizer:
The other domains are structured slightly differently. Since we need more than one service in our domain, then we should have a folder for each service from the start.

I think we should go with this structure for now:

```
domains/certificates/
├── .devcontainer
├── chart
├── Query.API/
│   ├── API
│   ├── Tests.Unit
│   └── Tests.App
├── Issuer/
│   ├── Worker
│   ├── Tests.Unit
│   └── Tests.App
└── certificates.sln
 ```

In your context you only need `Query.API` - and for now you can ignore the split between API (src) and Tests. So I am just suggesting to add a folder called "Query.API" and move the "hello world" service into that.

File: domains/certificates/Properties/launchSettings.json [hunk 1]
```diff
@@ -0,0 +1,31 @@
+﻿{
```
duizer, line 1: The Properties folder should be under Query.API

domains/certificates/Certificates.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
```
duizer, line 1: Move this to be under Query.API

duizer (approved):

PR #142, opened 2022-10-19
Certificates Query.Api mock

duizer:
There is no authentication enabled for this API. As it is a mock API it is technically not needed, but it will be needed later. The normal way of doing this is through our custom nuget library, but I would like to use the authentication library built-in to asp.net core. In order to test if the built-in authentication library works we must be able to deploy the API to an environment, which is currently not possible.

### DIFF
--- a/domains/certificates/Query.API/API/Controllers/CertificatesController.cs
+++ b/domains/certificates/Query.API/API/Controllers/CertificatesController.cs
@@ -9,7 +9,7 @@ namespace API.Controllers;
 public class CertificatesController : ControllerBase
 {
     [HttpGet]
-    [Route("v1/certificates")]
+    [Route("certificates")]
     public ActionResult<CertificateList> Get()
     {
         var gsrns = new[]
--- a/domains/certificates/Query.API/API.AppTests/ApiTests.cs
+++ b/domains/certificates/Query.API/API.AppTests/ApiTests.cs
@@ -18,7 +18,7 @@ public class ApiTests : IClassFixture<QueryApiWebApplicationFactory>
     public async Task GetList_AppStarted_ReturnsCertificates()
     {
         var client = factory.CreateClient();
-        var certificatesResponse = await client.GetFromJsonAsync<CertificateList>("v1/certificates");
+        var certificatesResponse = await client.GetFromJsonAsync<CertificateList>("certificates");
 
         const int numberOfMeteringPoints = 3;
         const int numberOfHours = 24;
--- a/domains/certificates/Query.API/API.AppTests/SwaggerTests.GetSwaggerDoc_AppStarted_NoChangesAccordingToSnapshot.verified.txt
+++ b/domains/certificates/Query.API/API.AppTests/SwaggerTests.GetSwaggerDoc_AppStarted_NoChangesAccordingToSnapshot.verified.txt
@@ -5,7 +5,7 @@
     "version": "v1"
   },
   "paths": {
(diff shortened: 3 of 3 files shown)


domains/certificates/Query.API/API/Models/Certificate.cs
```diff
@@ -0,0 +1,24 @@
...
+
+    /// <summary>
+    /// Amount of energy measured in Wh
+    /// </summary>
+    public int Amount { get; set; }
+
+    /// <summary>
+    /// Id for the metering point
+    /// </summary>
+    public string MeteringPointId { get; set; }
```
MartinSchmidt, line 23: We have used other naming on the measurements and emissions api, i would recommend we use the same wording:
Also Long for quantity, since we have previously have hit the limit of an int.
https://github.com/Energinet-DataHub/energy-origin/blob/main/domains/measurements/Measurements.API/API/Models/Measurement.cs


    string GSRN,
    long DateFrom,
    long DateTo,
    long Quantity

MartinSchmidt (changes requested):

domains/certificates/Query.API/API/Models/Certificate.cs
duizer: Regarding long for quantity, I have a question and would like your input: I considered long, but when we hit the limit that was related to adding many measurements together. Since this is a single certificate for a single metering point with a short time period, the quantity is not going to be above 2,147,483,647 Wh (+2 TWh). I do not mind using long in this case, so it is consistent with the remaining API, but in this case with a dedicated read/query model we can use a smaller data type. So do we want API consistency or smaller data footprint?

Regarding usage of GSRN: Do we want to keep calling the metering point id for "GSRN"? I think I heard something about moving away from calling it "GSRN" as it is a standard used in Denmark and not many other countries.

MartinSchmidt: Regarding long, I would agree, but we will most likely create a API bucketed on time, where it would be long, and to remove possible confusion about when to use int and when to use long in the future, json does not have the idea of longs, so it does not change anything there, it is only internally related.

GSRN is a global standard where we have bought a series: https://www.gs1si.org/Portals/0/GS1_Dokumentacija/GS1_Publikacije/Brosure_GS1_ANG/Introduction_GSRN.pdf so it is not a danish thing.

Exobitt (dismissed):

domains/certificates/Query.API/API/Controllers/CertificatesController.cs
```diff
@@ -0,0 +1,39 @@
...
+using API.Models;
+using Microsoft.AspNetCore.Mvc;
+
+namespace API.Controllers;
+
+[ApiController]
+public class CertificatesController : ControllerBase
+{
+    [HttpGet]
+    [Route("v1/certificates")]
```
Exobitt, line 12: Are we sure about this route?

duizer: We have not discussed/designed the API upfront. At this point we only need a single endpoint/route, so I thought this was the simplest and also gave some versioning.

duizer: After talking with Danni we found that having "v1" in the route will be strange wrt. the ingressroute. Since API versioning is not in place in general on Energioprindelse, we will remove it.

Exobitt (approved):

PR #144, opened 2022-10-20
Remove EnergyOriginDateTimeExtension lib from eventstore

duizer:
There is no need for the EnergyOriginDateTimeExtension. Furthermore, since EnergyOriginDateTimeExtension uses DateTime there is a risk that other developers will use DateTime when they should be using DateTimeOffset or Unix time instead. 
We need to get rid of it and let's start with this central component.


duizer:
Dropping this as we are dropping event store.

PR #146, opened 2022-10-21
Plan out domain workflow

CodeReaper:


PeterAGY (dismissed):
Looks good to me

Exobitt (approved):

.github/workflows/build-domain.yaml
```diff
@@ -0,0 +1,288 @@
...
+          find domains -mindepth 1 -maxdepth 1 -type d | while read domain; do
+            required_folder "$domain/.devcontainer"
+            required_file "$domain/.devcontainer/Dockerfile"
+            required_file "$domain/.devcontainer/devcontainer.json"
+
+            required_folder "$domain/.vscode"
+            required_file "$domain/.vscode/settings.json"
+            required_file "$domain/.vscode/extensions.json"
+
+            required_file "$domain"/*.sln
```
duizer, line 43: This is just a summary of what we talked about in person: These files and folders are technically not required for building the domain. They are listed here as required to ensure that each domain provide a working setup for development.
@CodeReaper: Please mark this conversation as resolved if you agree.

duizer (changes requested):

.github/workflows/build-domain.yaml
```diff
@@ -0,0 +1,292 @@
...
+
+          if [ -z "$pattern" ]; then
+            echo 'Relevance pattern is empty, emitting empty matrices.'
+            echo '::set-output name=domains::[]'
+            echo '::set-output name=projects::[]'
+            exit 0
+          fi
+
+          matrix=$(find domains -mindepth 1 -maxdepth 1 -type d | egrep "$pattern" | while read domain; do
+            projects=$(find "$domain" -mindepth 3 -maxdepth 3 -name "*.csproj" -exec dirname {} \; | egrep -v '[^\/]*/[^\/]*/Shared/[^\/]*' | egrep -v 'Tests$' | while read directory; do printf '"%s"' "$directory"; done | jq -sc '.')
```
duizer, line 102: (Hope that the following makes sense and I am using the correct terms regarding "service" and "project")
Maybe this is something to be done at a later point, but what if we have a "service" that is only tests? An example could be BDD/Specification tests (which Team Atlas is looking into right now). These kinds of tests are not extra projects to a main project for a service - that is that it is not a test of Service A or Service B, but on a higher level.

For those kinds of tests, I would expect them to be executed on every change within the domain.

.github/workflows/build-domain.yaml
```diff
@@ -0,0 +1,292 @@
...
+            required_file "$domain/.vscode/settings.json"
+            required_file "$domain/.vscode/extensions.json"
+
+            required_file "$domain"/*.sln
+
+            required_folder "$domain/chart"
+            required_file "$domain/chart/Chart.yaml"
+            required_file "$domain/chart/values.yaml"
+
+            find "$domain" -mindepth 3 -maxdepth 3 -name "*.csproj" -exec dirname {} \; | egrep -v '[^\/]*/[^\/]*/Shared/[^\/]*' | egrep -v 'Tests$' | while read project; do
```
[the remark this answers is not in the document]
CodeReaper: The relevance pattern operates on a domain basis, i.e. if there is any change within a single domain every project is built.

Building only the effected projects is an optimization we have discussed, but have not attempted to implement.

Please resolve or elaborate :)

.github/workflows/build-domain.yaml
CodeReaper: We could include a dummy project to ensure the tests runs, but that would also publish a dummy service.

Is there a reason why these sets of tests cannot be located within an existing project? (Since that would make them run with every change within the domain as-is)

duizer (dismissed):

NielsJorck (approved):
LGTM 🐑

PR #147, opened 2022-10-24
Preview/919 skeleton

duizer:


domains/certificates/Issuer/Issuer.Worker/DataSyncSyncer/Startup.cs
```diff
@@ -0,0 +1,11 @@
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Issuer.Worker.DataSyncSyncer;
+
+public static class Startup
```
ckr123, line 5: What is the difference between `Startup` and `Program`?

ckr123 (approved):

PR #150, opened 2022-10-28
956-domain-specific-editor

ckr123:


domains/certificates/.editorconfig
```diff
@@ -0,0 +1,128 @@
+###############################
+# Core EditorConfig Options   #
+###############################
+root=true
```
duizer, line 4: Do we really want this to be root=true? I would be fine with just overwriting the parts of editorconfig that we want.

PR #155, opened 2022-11-04
Start frontend integration

MadsThy:


MadsThy:
Will do it differently

PR #158, opened 2022-11-08
Feature/907 implement datasync

ckr123:


Exobitt (approved):

domains/certificates/Query.API/API/DataSyncSyncer/Service/DataSync.cs
```diff
@@ -0,0 +1,71 @@
...
+        this.options = options.Value;
+    }
+
+    public async Task<EnergyMeasuredIntegrationEvent> GetMeasurement(string gsrn, Period period)
+    {
+        logger.LogInformation(
+            "Fetching data in period from {from} to: {to}", period.DateFrom, period.DateTo
+        );
+
+        var url = $"{options.Url}/measurements?gsrn={gsrn}&dateFrom={period.DateFrom}&dateTo={period.DateTo}";
```
duizer, line 40: It is also possible to set the BaseAddress on the HttpClient. Then you do not have to prepend it here. Furthermore, this will also fix if someone in the config ends the url with a "/"

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncSyncerWorker.cs
```diff
@@ -13,41 +18,58 @@ namespace API.DataSyncSyncer;
...
 
         while (!stoppingToken.IsCancellationRequested)
         {
-            if (!string.IsNullOrWhiteSpace(gsrn))
+            var period = GetPeriod();
+
+            foreach (var data in masterData.Where(it => !string.IsNullOrWhiteSpace(it.GSRN)))
             {
+                var measurement = await dataSync.GetMeasurement(data.GSRN, period);
+                await integrationEventBus.Produce(stoppingToken, measurement);
```
duizer, line 48: I think this principle will not work in production. On each "timer tick" we are asking for measurements from the last hour. But I do not think DataHub will have data ready that fast. 
So the principle should more be that it fetches data from the last time that metering point received measurements on each "timer tick".

duizer (changes requested):

PR #161, opened 2022-11-09
Preview/datasync impl

ckr123:


duizer (changes requested):

domains/certificates/Query.API/API/appsettings.Development.json
```diff
@@ -7,5 +7,8 @@
     },
     "MockMasterData": {
         "JsonFilePath": "MasterDataService/data.json"
+    },
+    "Datasync": {
+        "url": "http://somecoolurl:8081"
```
duizer, line 12:
```suggestion
        "Url": "http://somecoolurl:8081"
```

domains/certificates/Query.API/API/DataSyncSyncer/Service/DataSync.cs
```diff
@@ -0,0 +1,67 @@
...
+    public DataSync(HttpClient httpClient, ILogger<DataSync> logger)
+    {
+        this.logger = logger;
+        this.httpClient = httpClient;
+    }
+
+    public async Task<List<DataSyncDto>> GetMeasurement(string gsrn, Period period, string meteringPointOwner)
+    {
+        logger.LogInformation(
+            "Fetching data in period from {from} to: {to}", period.DateFrom, period.DateTo
```
duizer, line 30: Suggestion: This will print epoch time, so I suggest to write it in a more friendly format

domains/certificates/Query.API/API/DataSyncSyncer/Service/DataSync.cs
```diff
@@ -0,0 +1,67 @@
...
+        );
+
+        var url = $"measurements?gsrn={gsrn}&dateFrom={period.DateFrom}&dateTo={period.DateTo}";
+
+        httpClient.SetBearerToken(GenerateToken(meteringPointOwner));
+
+        var response = await httpClient.GetAsync(url);
+
+        return await response.Content.ReadFromJsonAsync<List<DataSyncDto>>()
+               ?? throw new Exception($"Fetch of measurements failed, base: {httpClient.BaseAddress} url: {url}");
```
duizer, line 40: If there is a failure, then this will throw an exception. I do not see something that will catch this exception. So a single exception will bring down the entire DataSyncSyncerWorker.
I suggest that we here will not throw an exception if DataSync is not responding, but will return either an empty result or a type indicating that something went wrong.

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncSyncerWorker.cs
```diff
@@ -13,41 +17,94 @@ namespace API.DataSyncSyncer;
...
         }
+    }
 
-        await Task.Delay(TimeSpan.FromMilliseconds(500),
-            stoppingToken); //allow application to start before producing events
-        var random = new Random();
+    private async Task SleepToNearestHour(CancellationToken stoppingToken)
+    {
+        var minutesToNextHour = 60 - DateTimeOffset.Now.Minute;
+        logger.LogInformation("Minutes to next hour {minutesToNextHour}", minutesToNextHour);
```
duizer, line 62: Suggestion to make it a bit easier to read the logs:

```suggestion
        logger.LogInformation("Sleeping until next full hour ({minutesToNextHour} minutes)", minutesToNextHour);
```

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncSyncerWorker.cs
```diff
@@ -13,41 +17,94 @@ namespace API.DataSyncSyncer;
...
+            await SleepToNearestHour(stoppingToken);
+
+            foreach (var data in masterData)
+            {
+                var measurements = await FetchMeasurements(data.GSRN, data.MeteringPointOwner);
+                SetNextPeriodStartTime(measurements, data.GSRN);
+
+                var integrationsEvents = MapToIntegrationEvents(measurements);
+                await integrationEventBus.Produce(stoppingToken, integrationsEvents);
+                logger.LogInformation("Produce energy measured event");
```
duizer, line 54: This logging statement is not helpful for us in a debugging scenario. It could be either be deleted or changed to 

```suggestion
                logger.LogInformation("Produced {numEvents} energy measured events", integrationEvents.Count);
```

so we from the logs know if it is actually producing/publishing events to the bus.

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncSyncerWorker.cs
```diff
@@ -13,41 +17,94 @@ namespace API.DataSyncSyncer;
...
     {
-        if (string.IsNullOrWhiteSpace(gsrn))
+        while (!stoppingToken.IsCancellationRequested)
         {
-            logger.LogWarning("No master data loaded. Will not produce any events");
+            await SleepToNearestHour(stoppingToken);
+
+            foreach (var data in masterData)
+            {
+                var measurements = await FetchMeasurements(data.GSRN, data.MeteringPointOwner);
```
duizer, line 49: The stoppingToken should be passed here to ensure that calls to DataSync will be cancelled as fast as possible in case the stoppingToken signals a cancellation.

duizer (changes requested):

PR #168, opened 2022-11-15
Add tech and fuel code to certificates API

duizer:


domains/certificates/Query.API/API/Query.API/ApiModels/Certificate.cs
```diff
@@ -24,4 +24,15 @@ public class Certificate
     /// </summary>
     [JsonPropertyName("gsrn")]
     public string GSRN { get; set; } = "";
+
```
duizer, line 27: After talking with Mads, we also need to add a certificate ID. Use string for now.

PR #171, opened 2022-11-17
preview-907-datasyync-impl2

ckr123:


domains/certificates/Query.API/API.UnitTests/DataSyncSyncer/DataSyncServiceTest.cs
```diff
@@ -0,0 +1,164 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using API.DataSyncSyncer.Dto;
+using API.MasterDataService;
+using CertificateEvents.Primitives;
+using IntegrationEvents;
+using Microsoft.Extensions.Logging;
+using Moq;
+using Newtonsoft.Json;
+using RichardSzalay.MockHttp;
```
Exobitt, line 15: Why use both Moq and MockHttp?

domains/certificates/chart/values.yaml
```diff
@@ -1,6 +1,9 @@
 api:
   replicaCount: 1
   image: {}
+  env:
+    MOCKMASTERDATA__JSONFILEPATH:
+    DATASYNC__URL:
```
Exobitt, line 6: These should be removed, as eo-base will overwrite them. Therfore, they should be added over there

domains/certificates/Query.API/API/MasterDataService/data.json
```diff
@@ -8,6 +8,6 @@
             "techCode": "T010000"
         },
         "meteringPointOwner": "280d0d0e-acde-4132-9916-c2299f583325",
-        "meteringPointOnboarded": true
+        "meteringPointOnboarded": "2022-09-11T19:13:50+00:00"
```
Exobitt, line 11: Is is correct, to move from a boolean to datetime?
If so, would it make sense to rename the field?

domains/certificates/Query.API/API/DataSyncSyncer/Startup.cs
```diff
@@ -1,9 +1,18 @@
+using System;
```
Exobitt, line 1: Can all in Startup.cs be moved into Program.cs?

domains/certificates/Query.API/API.UnitTests/DataSyncSyncer/DataSyncServiceTest.cs
ckr123: Moq is used to mock the DI classes and mockHttp is used to mock the httpClient.
as described here https://github.com/Energinet-DataHub/energy-origin/blob/main/doc/architecture/adr/0010-nuget-mockhttp.md

domains/certificates/Query.API/API/DataSyncSyncer/Startup.cs
ckr123: We have a separate Startup in each component to separate the different services and easily replace them when needed.

domains/certificates/Query.API/API/MasterDataService/data.json
ckr123: Yes we need it to know when to start fetching data. This might be changed when datasync is properly implemented by heimdal.
renamed in 828ac08

domains/certificates/chart/values.yaml
ckr123: removed in a893442

domains/certificates/Query.API/API/MasterDataService/data.json
duizer: It is desired to have it as a timestamp. This allows us to say that certificates should be generated for a metering point starting at a specific timestamp - and the timestamp can be in the future (e.g. Jan 1st 2023)

I agree that it makes sense to rename the property. I suggest "issuanceStartDate"

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncService.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Threading.Tasks;
+using API.DataSyncSyncer.Dto;
+using CertificateEvents.Primitives;
+using Marten;
+using Microsoft.Extensions.Logging;
+using Microsoft.IdentityModel.Tokens;
+
+namespace API.DataSyncSyncer;
+
+public class DataSyncService
```
duizer, line 20: I see this class as having multiple responsibilities: 

1) Updating the sync state (periodStartTimeDictionary)
2) Logic about a creating a period to request measurements from and if the period is valid
3) Calling DataSync

Given that we will add storage for persisting the sync state, then it would make sense to extract a class/interface for this storage. This will also remove some of the responsibility from this class

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncService.cs
```diff
@@ -0,0 +1,123 @@
...
+        periodStartTimeDictionary = state;
+    }
+
+    public async Task<List<DataSyncDto>> FetchMeasurements(string GSRN, string meteringPointOwner,
+        CancellationToken cancellationToken)
+    {
+        var dateFrom = periodStartTimeDictionary[GSRN].ToUnixTimeSeconds();
+
+        var now = DateTimeOffset.UtcNow;
+        var midnight = new DateTimeOffset(now.Year, now.Month, now.Day, 0, 0, 0, TimeSpan.Zero).ToUnixTimeSeconds();
```
duizer, line 43: Suggestion: To increase readability we could create an extension method on DateTimeOffset that returns midnight, so we can have something like:

```
var midnight = DateTimeOffset.UtcNow.GetMidnight();
```

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncService.cs
```diff
@@ -0,0 +1,123 @@
...
+                        DateFrom: dateFrom,
+                        DateTo: midnight
+                    ),
+                    meteringPointOwner,
+                    cancellationToken
+                );
+            }
+            catch (Exception e)
+            {
+                logger.LogInformation("An error occured: {error}, no measurements was fetched", e.Message);
```
duizer, line 62:
```suggestion
                logger.LogInformation("An error occured: {error}, no measurements were fetched", e.Message);
```

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncService.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        var newestMeasurement = measurements.Max(m => m.DateTo);
+        periodStartTimeDictionary![GSRN] = DateTimeOffset.FromUnixTimeSeconds(newestMeasurement);
+    }
+
+    private async Task<List<DataSyncDto>> RequestDatasync(string gsrn, Period period, string meteringPointOwner,
+        CancellationToken cancellationToken)
+    {
+        logger.LogInformation(
+            "Fetching data in period from {from} to: {to}", period.DateFrom.ToString("MM/dd/yy H:mm:ss"), period.DateTo.ToString("MM/dd/yy H:mm:ss")
```
duizer, line 86: `Perdod.DateFrom` and `Period.DateTo` are of type "long" and this will not print a date.

This: `DateTimeOffset.UtcNow.ToUnixTimeSeconds().ToString("MM/dd/yy H:mm:ss")` will return the string "MM/dd/yy H:mm:ss"

If you want to print the timestamp as something to easily read, then you have to parse it back to DateTimeOffset and then call `.ToString("o")` on that.

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncService.cs
```diff
@@ -0,0 +1,123 @@
...
+        );
+
+        var url = $"measurements?gsrn={gsrn}&dateFrom={period.DateFrom}&dateTo={period.DateTo}";
+        httpClient.DefaultRequestHeaders.Authorization =
+            new AuthenticationHeaderValue("Bearer", GenerateToken(meteringPointOwner));
+
+        var response = await httpClient.GetAsync(url, cancellationToken);
+
+        return await response.Content.ReadFromJsonAsync<List<DataSyncDto>>(cancellationToken: cancellationToken)
+               ?? throw new Exception($"Fetch of measurements failed, base: {httpClient.BaseAddress} url: {url}");
```
duizer, line 96: Why throw this exception here? I am not sure when this exception will be reached. If the call returns a status code different than 200, then line 95 will result in an exception because there is no content (or the content is definitely not Json with List<DataSyncDto>).

Maybe it make more sense to call explicitly https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpresponsemessage.ensuresuccessstatuscode?view=net-6.0 before trying to read the content.

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncService.cs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    private static string GenerateToken(string meteringPointOwner)
+    {
+        var claims = new Claim[]
+        {
+            new(JwtRegisteredClaimNames.UniqueName, "username"),
+            new(JwtRegisteredClaimNames.NameId, Guid.NewGuid().ToString()),
```
duizer, line 104: This is the payload for a JWT from the auth service:

```
{
  "issued": "2022-11-17T06:46:29.147835+00:00",
  "expires": "2022-11-18T06:46:29.147835+00:00",
  "actor": "7f497b4a-8bfd-4ed4-a12f-6ab9235d998b",
  "subject": "c2410975-fa5f-441d-a40e-6231ed540495",
  "scope": [
    "meteringpoints.read",
    "measurements.read"
  ]
}
```

I think we should not add more claims than what is currently in the JWT token from the auth service, so I think we should remove these two claims

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncService.cs
```diff
@@ -0,0 +1,123 @@
...
+            new("subject", meteringPointOwner),
+            new("actor", "actor"),
+            new("scope", "scope"),
```
duizer, line 107: These do not have a valid value. We could change it so, "actor" is the same value as "subject" and then have the scopes "meteringpoints.read" and"measurements.read". To add multiple scopes I believe you have to make multiple Claim with the same key ("scope")

domains/certificates/Query.API/API/DataSyncSyncer/Startup.cs
```diff
@@ -1,9 +1,18 @@
...
 public static class Startup
 {
-    public static void AddDataSyncSyncer(this IServiceCollection services)
-        => services.AddHostedService<DataSyncSyncerWorker>();
+    public static void AddDataSyncSyncer(this IServiceCollection services, IConfiguration configuration)
+    {
+        var datasyncUrl = configuration.GetSection(DatasyncOptions.Datasync).Value;
+        services.AddHttpClient<DataSyncService>(client => client.BaseAddress = new Uri(datasyncUrl));
+
+        services.AddTransient<DataSyncService>();
```
duizer, line 15: Just a note: DataSyncService is lifetime "transient". If DataSyncService holds state, then that lifetime is a problem.

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncSyncerWorker.cs
```diff
@@ -14,36 +18,64 @@ internal class DataSyncSyncerWorker : BackgroundService
...
 
-            logger.LogInformation("Publish EnergyMeasuredIntegrationEvent");
+            foreach (var data in masterData)
+            {
+                var measurements = await dataSyncService.FetchMeasurements(data.GSRN, data.MeteringPointOwner,
+                    cancellationToken);
 
-            await Task.Delay(TimeSpan.FromSeconds(10), stoppingToken);
+                var integrationsEvents = MapToIntegrationEvents(measurements);
+                await bus.Publish(integrationsEvents, cancellationToken);
```
duizer, line 56: What will this do in case of no integrationsEvents? Will it do nothing? Will it throw an exception?

domains/certificates/Query.API/API/GranularCertificateIssuer/EnergyMeasuredConsumer.cs
```diff
@@ -60,13 +60,14 @@ public async Task Consume(ConsumeContext<EnergyMeasuredIntegrationEvent> context
 
     private static bool ShouldEventBeProduced(MasterData? masterData)
     {
+        var now = DateTime.Now;
```
duizer, line 63:
```suggestion
        var now = DateTimeOffset.UtcNow;
```

Technically not needed in this case, but as a default I would go with using DateTimeOffset over DateTime as I am then always sure to have a single point in time.

domains/certificates/Query.API/API.UnitTests/GranularCertificateIssuer/EnergyMeasuredConsumerTest.cs
```diff
@@ -70,7 +70,7 @@ public async Task Consume_ConsumptionPoint_NoEventsSaved()
     [Fact]
     public async Task Consume_MeteringPointNotOnboarded_NoEventsSaved()
```
duizer, line 71:
```suggestion
    public async Task Consume_OneDayBeforeIssuranceStartDate_NoEventsSaved()
```

domains/certificates/Query.API/API.UnitTests/GranularCertificateIssuer/EnergyMeasuredConsumerTest.cs
```diff
@@ -24,7 +24,7 @@ public class EnergyMeasuredConsumerTest
         Type: MeteringPointType.Production,
         Technology: new Technology(FuelCode: "F00000000", TechCode: "T010000"),
         MeteringPointOwner: "meteringPointOwner",
-        MeteringPointOnboarded: true);
+        MeteringPointOnboardedStartDate: DateTimeOffset.Now.AddDays(-1));
```
duizer, line 27: We could add tests where 
- we are just before (e.g. 1 ms before) the issuance start date
- exactly on the issuance start date

This will require that we can control the time in EnergyMeasuredConsumer, but that can be done with an interface.
What do you think?

domains/certificates/Query.API/API.UnitTests/DataSyncSyncer/DataSyncServiceTest.cs
```diff
@@ -0,0 +1,163 @@
...
+        fakeHttpClient.VerifyNoOutstandingExpectation();
+    }
+
+    [Fact]
+    public async Task GetMeasurements_MeteringPointOnboarded_DateFromUpdated()
+    {
+        var meteringPointOnboarded = DateTimeOffset.Now.AddDays(-2);
+        var service = SetupService(meteringPointOnboarded);
+
+        await AssertDateFromQueryParamIsUpdated(meteringPointOnboarded, service);
```
duizer, line 65: Please inline this method

domains/certificates/Query.API/API.UnitTests/DataSyncSyncer/DataSyncServiceTest.cs
```diff
@@ -0,0 +1,163 @@
...
+        client.BaseAddress = new Uri("http://localhost:8080");
+
+        var masterDataList = new List<MasterData>
+        {
+            validMasterData with
+            {
+                MeteringPointOnboardedStartDate = meteringPointOnboardedStartDate
+            }
+        };
+
```
duizer, line 149: This space could be removed

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncService.cs
duizer: and then add "issued" and "expires" claims. Lifetime validation is not performed by DataSync, but then we are doing the same as AuthService is currently doing

domains/certificates/Query.API/API.UnitTests/GranularCertificateIssuer/EnergyMeasuredConsumerTest.cs
ckr123: Is your suggestion to inject MockMasterDataCollection into SyncState?

domains/certificates/Query.API/API.UnitTests/GranularCertificateIssuer/EnergyMeasuredConsumerTest.cs
```diff
@@ -94,12 +102,11 @@ public async Task Consume_ProductionPoint_EventsSaved()
         var masterDataServiceMock = new Mock<IMasterDataService>();
         masterDataServiceMock.Setup(m => m.GetMasterData(It.IsAny<string>())).ReturnsAsync(validMasterData);
```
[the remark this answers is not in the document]
ckr123: If we set it to now, there won't be any event saved in this test

domains/certificates/Query.API/API/DataSyncSyncer/Startup.cs
```diff
@@ -1,9 +1,23 @@
...
 namespace API.DataSyncSyncer;
 
 public static class Startup
 {
-    public static void AddDataSyncSyncer(this IServiceCollection services)
-        => services.AddHostedService<DataSyncSyncerWorker>();
+    public static void AddDataSyncSyncer(this IServiceCollection services, IConfiguration configuration)
+    {
+        var datasyncUrl = configuration.GetSection(DatasyncOptions.Datasync).Value;
+        services.AddHttpClient<DataSyncService>(client => client.BaseAddress = new Uri(datasyncUrl));
```
duizer, line 15: This does not correctly bind an instance of DatasyncOptions. This works:

```suggestion
        DatasyncOptions options = new();
        configuration.GetRequiredSection(DatasyncOptions.Datasync).Bind(options);
        services.AddHttpClient<DataSyncService>(client => client.BaseAddress = new Uri(options.Url));
```

duizer (approved):

PR #176, opened 2022-11-24
DataSyncSyncer uses Marten for sync period start 

duizer:


domains/certificates/Query.API/API/DataSyncSyncer/Persistence/SyncState.cs
```diff
@@ -1,33 +1,46 @@
...
+                .DefaultIfEmpty(0)
+                .Max();
+
+            return Math.Max(maxDateTo, masterData.MeteringPointOnboardedStartDate.ToUnixTimeSeconds());
+        }
+        catch (Exception e)
         {
-            return periodStart.ToUnixTimeSeconds();
+            logger.LogWarning("Failed reading from database. Exception: {exception}", e);
+            return null;
```
Exobitt, line 43: Should this return null instead of raising an exception?

domains/certificates/Query.API/API/DataSyncSyncer/Persistence/SyncState.cs
```diff
@@ -1,33 +1,46 @@
...
+        this.logger = logger;
     }
 
-    public long GetPeriodStartTime(string GSRN, DateTimeOffset meteringPointOnboardedStartDate)
+    public async Task<long?> GetPeriodStartTime(MasterData masterData)
     {
-        if (periodStartTimeDictionary.TryGetValue(GSRN, out var periodStart))
+        try
+        {
+            await using var querySession = documentStore.QuerySession();
```
ckr123, line 26: I previously had problems with inline using not disposing as expected in https://github.com/Energinet-DataHub/energy-origin/blob/main/domains/auth/Auth.API/API/Utilities/Cryptography.cs#L31

Is this relevant here?

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncService.cs
```diff
@@ -29,20 +28,25 @@ public DataSyncService(IDataSyncClient client, ILogger<DataSyncService> logger,
     public async Task<List<DataSyncDto>> FetchMeasurements(MasterData masterData,
         CancellationToken cancellationToken)
     {
-        var dateFrom = syncState.GetPeriodStartTime(masterData.GSRN, masterData.MeteringPointOnboardedStartDate);
+        var dateFrom = await syncState.GetPeriodStartTime(masterData);
+
+        if (dateFrom == null)
+        {
+            logger.LogInformation("Not possible to get dateFrom from sync state for {masterData}", masterData);
```
ckr123, line 35: suggestion 
```suggestion
            logger.LogInformation("Not possible to get start date from sync state for {masterData}", masterData);
```

domains/certificates/Query.API/API.UnitTests/DataSyncSyncer/Persistence/SyncStateTest.cs
```diff
@@ -1,48 +1,145 @@
...
 
     [Fact]
-    public void GetPeriodStartTime_GetTime()
+    public async Task GetPeriodStartTime_OneCertificateInStore_ReturnsMeteringPointOnboardedStartDate()
     {
-        var now = DateTimeOffset.Now;
-        var currentStartTime = syncState.GetPeriodStartTime("gsrn", now);
-        Assert.Equal(now.ToUnixTimeSeconds(), currentStartTime);
+        var view = new CertificatesByOwnerView { Owner = masterData.MeteringPointOwner };
+        var start = masterData.MeteringPointOnboardedStartDate;
```
ckr123, line 43: Could be nice with some new lines for readability like in `GetPeriodStartTime_TwoCertificatesInStore_ReturnsDataMaxDate()`

domains/certificates/Query.API/API.UnitTests/DataSyncSyncer/Persistence/SyncStateTest.cs
```diff
@@ -1,48 +1,145 @@
...
+        var actualPeriodStartTime = await syncState.GetPeriodStartTime(masterData);
+
+        actualPeriodStartTime.Should().Be(cert2DateTo);
+    }
+
+    [Fact]
+    public async Task GetPeriodStartTime_NoCertificatesForThatGsrnInStore_ReturnsMeteringPointOnboardedStartDate()
+    {
+        var view = new CertificatesByOwnerView { Owner = masterData.MeteringPointOwner };
+        var start = masterData.MeteringPointOnboardedStartDate;
```
ckr123, line 103: I would also suggest some new lines here

domains/certificates/Query.API/API.UnitTests/DataSyncSyncer/DataSyncServiceTest.cs
```diff
@@ -88,7 +88,7 @@ public async Task FetchMeasurements_MeteringPointOnboarded_NoDataFetched()
         var response = await service.FetchMeasurements(masterData,
             CancellationToken.None);
 
-        Assert.Empty(response);
+        response.Should().BeEmpty();
```
ckr123, line 91: Should we agree on always using FluentAssertions in our tests?

domains/certificates/Query.API/API/DataSyncSyncer/Persistence/SyncState.cs
duizer: This is running in a background service and an unhandled exception will stop the background service and the application. If we throw an exception here, then something outside needs to handle it. 

I think it is better to catch this at the lower levels. That is why it can return null.

domains/certificates/Query.API/API/DataSyncSyncer/Persistence/SyncState.cs
duizer: It is not the same situation since we are not dealing with streams. Here the using-statement is used to ensure that the session is closed when exiting the method. In Cryptography.cs the using statement would cause all buffered data to be written to the underlying stream when exiting the method; but that would be too late in Cryptography.cs and that is why an explicit call to `.Close()` was added.

domains/certificates/Query.API/API.UnitTests/DataSyncSyncer/Persistence/SyncStateTest.cs
duizer: fixed in 81f2db9fe13461baa97b115a32b88811f11d1a83

domains/certificates/Query.API/API.UnitTests/DataSyncSyncer/Persistence/SyncStateTest.cs
duizer: fixed

domains/certificates/Query.API/API.UnitTests/DataSyncSyncer/DataSyncServiceTest.cs
duizer: I would like that :) 
I prefer FluentAssertions as they are easier to read and makes testing for equality between objects easy

ckr123 (approved):

PR #177, opened 2022-11-24
Feature/1077 add more logging

ckr123:


domains/certificates/Query.API/API/DataSyncSyncer/DataSyncService.cs
```diff
@@ -48,6 +48,12 @@ public DataSyncService(IDataSyncClient client, ILogger<DataSyncService> logger,
                     masterData.MeteringPointOwner,
                     cancellationToken
                 );
+                logger.LogInformation(
+                    "Successfully fetched {numberOfMeasurements} measurements in period from {from} to: {to}",
```
duizer, line 52: Please add GSRN here. That is useful when we have more than 1 metering point.

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncService.cs
```diff
@@ -48,6 +48,12 @@ public DataSyncService(IDataSyncClient client, ILogger<DataSyncService> logger,
...
+                    result.Count,
+                        DateTimeOffset.FromUnixTimeSeconds(dateFrom).ToString("o"),
```
duizer, line 54: Can this be indented on the same level as the other arguments?

duizer (changes requested):

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncService.cs
ckr123: Fixed in e961e28

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncService.cs
ckr123: fixed in e961e28

duizer (approved):

PR #181, opened 2022-11-28
1041 persistent volume on postgres

Exobitt:


CodeReaper (approved):

domains/certificates/chart/templates/postgres-statefulset.yaml
```diff
@@ -15,29 +19,49 @@ spec:
...
+      volumes:
+        - name: dshm
+          emptyDir:
+            medium: Memory
+        - name: postgres-vm
+          configMap:
+            name: postgres-init-cm
+  volumeClaimTemplates:
+    - metadata:
+        name: data
```
duizer, line 61: @CodeReaper This is not referenced anywhere. Is it used? And if yes, where is it used?

PR #186, opened 2022-12-01
marten db in testcontainer

duizer:
https://github.com/JasperFx/marten is testing using containers. I have not been able to find an alternative using something like an in-memory or embedded or lite version. So I am using TestContainers for .NET (https://dotnet.testcontainers.org/). So far it is working nicely, but the downside is longer test execution time as a container with postgres must be started.

There has not been issues running this in Github actions. I have not tested this in the vscode dev container, but it is expected not to work as docker-in-docker is not setup for the vscode dev container. Everything else will work in the vs code container, but running the ApiTests will fail. There is a significant amount of time needed to solve this; but I think that we can live with this limitation for the devs using the dev container.

Reviewers are encouraged to run the test on their local setup to see if there are any issues.


domains/certificates/Query.API/API.AppTests/ApiTests.GetList_FiveMeasurementAddedToBus_ReturnsList.verified.txt
```diff
@@ -0,0 +1,54 @@
...
+      DateFrom: 1654056000,
+      DateTo: 1654059600,
+      Quantity: 46,
+      GSRN: Guid_2,
+      GridArea: GridArea,
+      TechCode: bar,
+      FuelCode: foo
+    },
+    {
+      Id: Guid_3,
```
Exobitt, line 14: Where is guid 2?

duizer: Verify.Xunit is scrubbing the data. So here it says that it expects a Guid and it labels that as "Guid_2", but it does not care about the specific value. In the file "Guid_2" is used for all GSRN properties, which means that it must be the same value.

It is nice in our case because the Id property is auto-generated by the code. And GSRN is auto-generated by the test.

duizer: But looking at it now, I think it would make sense to use a specific name for GSRN, like "123456" or "gsrn-1" since GSRN is not actually a Guid.

duizer: Changed in 7a01adc9e88d2871ca2647c4364ca8a1dd56fffd

Exobitt (approved):

PR #192, opened 2022-12-16
#1190: add sonarscanner to certificate domain

anna-knudsen:
Added .yaml file in .github/workflow to add sonarscanner to certificate domain.

Explanation:
[https://github.com/marketplace/actions/sonarcloud-scan](https://github.com/marketplace/actions/sonarcloud-scan)


anna-knudsen:
@cla-bot check

ckr123 (approved):

duizer:
The check fo SonarCloud quality gate failed, but it is possible to merge to main (the PR has been approved). As I remember if "Build domain" or "Check Versions" fails, then it is not possible to complete the PR. Can we have the same behaviour here for a failed sonarcloud quality gate?

duizer:
> The check fo SonarCloud quality gate failed, but it is possible to merge to main (the PR has been approved). As I remember, if "Build domain" or "Check Versions" fails, then it is not possible to complete the PR. Can we have the same behaviour here for a failed sonarcloud quality gate?

I remembered wrong. It is a setting on the repository to require checks to pass.

ckr123:
> The check fo SonarCloud quality gate failed, but it is possible to merge to main (the PR has been approved). As I remember, if "Build domain" or "Check Versions" fails, then it is not possible to complete the PR. Can we have the same behaviour here for a failed sonarcloud quality gate?

It is already beeing dealt with as one of the AC on the issues addresses this.

CodeReaper:
_My notes for the PR as it is now:_
- might need some configuration of concurrency to hopefully skip builds from intermediate commits when pushing individual commits.
- the scan seems to double-ish the time for checks to run, are there any configuration for sonar cloud that can reduce its run time?
- to have a required check in a mono repository it is important that a check with that name runs for every round of checks. So when using a `paths` filter you will need another workflow with a `paths-ignore` filter with the same name (see [stackoverflow](https://stackoverflow.com/a/70284299/190599)). It might be important to know where the name of checks comes from, so here is a [direct link](https://github.com/CodeReaper/find-diff-action/blob/d419926fd82f5740bf12212be34034233d3f941e/.github/workflows/test.yaml#L55) to the name of a check where the name is "Tests Succeeded".
- if there is a required check for pull requests, perhaps we do not really need run it again on pushes against the main branch.

I believe it would be beneficial to integrate the sonar cloud check directly into our `build-and-push-container` action where we can turn it on/off per build. We can add per-domain configuration on whether or not sonar cloud needs to run or not in the same manor as the draft property recently added to [auth domain](https://github.com/Energinet-DataHub/energy-origin/blob/main/domains/auth/.domain.json).
Advantages of integrating sonar cloud:
- Avoids build and testing twice and cut down on run time.
- build-domain and build-library workflows have the required checks.
- Concurrency is already handled

ckr123 (approved):

ckr123 (approved):

ckr123 (approved):

PR #204, opened 2023-01-06
Bugfix cache only valid owners masterdata

duizer:
`GetUuidForCompany` returns string.Empty when receving a not-ok answer from auth service. In this PR a check is added to ensure that MasterDataService only caches when there is a non empty string.

### DIFF
--- a/domains/certificates/Query.API/API/MasterDataService/MockMasterDataService.cs
+++ b/domains/certificates/Query.API/API/MasterDataService/MockMasterDataService.cs
@@ -58,7 +58,8 @@ internal class MockMasterDataService : IMasterDataService
 
         var meteringPointOwner = await clientFactory.CreateClient().GetUuidForCompany(cvr);
 
-        cvrToMeteringPointOwner.TryAdd(cvr, meteringPointOwner);
+        if (!string.IsNullOrWhiteSpace(meteringPointOwner))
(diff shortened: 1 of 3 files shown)


ckr123 (approved):

domains/certificates/Query.API/API/MasterDataService/MockMasterDataService.cs
```diff
@@ -58,7 +58,8 @@ private async Task<string> GetMeteringPointOwner(MasterDataMockInput masterDataM
 
         var meteringPointOwner = await clientFactory.CreateClient().GetUuidForCompany(cvr);
 
-        cvrToMeteringPointOwner.TryAdd(cvr, meteringPointOwner);
+        if (!string.IsNullOrWhiteSpace(meteringPointOwner))
```
ckr123, line 61: Shouldnt we have a test to catch this?

duizer: Since this is a mock implementation of the master data service and not the real implementation we decided not to add tests. However, the complexity of the mock implementation has increased since we started, which means we should revise our initial decision of no tests. 
I suggest this as a two step: Step 1: Get this bugfix out. Step 2: Add unit tests to MockMasterDataService.

ckr123 (approved):

ckr123 (approved):

PR #210, opened 2023-01-13
1242 validation for gc enabling

duizer:
The entity is either called `SignUp` or `CertificateGenerationSignUp`. This may lead to confusion. Let me know what you think. Please also note that the term has not been accepted (or defined) by the business at this point. 

The code smell reported by SonarCloud for roslyn:xUnit1033 is false (https://github.com/xunit/xunit/issues/2591). Please ignore for now


domains/certificates/Query.API/API/CertificateGenerationSignUpService/CertificateGenerationSignUpServiceImpl.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            return new GsrnNotFound();
+        }
+
+        if (matchingMeteringPoint.Type != MeterType.Production)
+        {
+            return new NotProductionMeteringPoint();
+        }
+
+        // Check if GSRN is already signed up
```
ckr123, line 39: If these comments are necessary I think this could be refactored to a method with a descriptive name so we can remove the comments

domains/certificates/Query.API/API/CertificateGenerationSignUpService/CertificateGenerationSignUpServiceImpl.cs
```diff
@@ -0,0 +1,76 @@
...
+        var document = await repository.GetByGsrn(gsrn);
+
+        if (document != null)
+        {
+            return new SignUpAlreadyExists(document);
+        }
+
+        // Save
```
ckr123, line 47: Could be deleted

domains/certificates/Query.API/API/CertificateGenerationSignUpService/CertificateGenerationSignUpServiceImpl.cs
```diff
@@ -0,0 +1,76 @@
...
+        this.client = client;
+        this.repository = repository;
+    }
+
+    public async Task<CreateSignUpResult> Create(string gsrn, string meteringPointOwner, DateTimeOffset startDate, CancellationToken cancellationToken)
+    {
+        var meteringPoints = await client.GetMeteringPoints(meteringPointOwner, cancellationToken);
+        var matchingMeteringPoint = meteringPoints?.MeteringPoints.FirstOrDefault(mp => mp.GSRN == gsrn);
+
+        if (matchingMeteringPoint == null)
```
ckr123, line 29: Could this be extracted so the create method doesnt handle the validation?

domains/certificates/Query.API/API/Query.API/Controllers/SignUpsController.cs
```diff
@@ -0,0 +1,101 @@
...
+using static API.CertificateGenerationSignUpService.CreateSignUpResult;
+
+namespace API.Query.API.Controllers;
+
+[Authorize]
+[ApiController]
+public class SignUpsController : ControllerBase
+{
+    /// <summary>
+    /// Signs up a metering point for granular certificate generation
```
ckr123, line 23: Are these summaries used in the opanapi doc?

duizer: Yes

domains/certificates/Query.API/API/CertificateGenerationSignUpService/CertificateGenerationSignUpServiceImpl.cs
duizer: Deleted

domains/certificates/Query.API/API/CertificateGenerationSignUpService/CertificateGenerationSignUpServiceImpl.cs
duizer: Deleted comments

Exobitt (approved):
lgtm

domains/certificates/Query.API/API.AppTests/SwaggerTests.GetSwaggerDoc_AppStarted_NoChangesAccordingToSnapshot.verified.txt
```diff
@@ -36,6 +36,137 @@
           }
         }
       }
+    },
+    "/api/signUps": {
```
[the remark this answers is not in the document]
duizer: It is based on the route defined using the [Route]-attribute on a controller (ContractsController in this case). So changing the [Route]-attribute will result in a changed Swagger/OpenAPI-doc

domains/certificates/Query.API/API/CertificateGenerationSignUpService/CertificateGenerationSignUpServiceImpl.cs
duizer: It could, but since the return type is a union type (or as close as we can do a union type), extracting a validation method will require that it returns a nullable result, which should then be checked in `Create`. So I prefer keeping it like this at this point.

PR #214, opened 2023-01-20
Validate StartDate is before year 10000

duizer:
If someone calls the API and uses Unix time in milliseconds and not seconds for `StartDate`, the code will fail when parsing to DateTimeOffset. This sets an upper limit on `StartDate`


Exobitt (dismissed):
lgtm

Exobitt (approved):
still lgtm

PR #225, opened 2023-02-09
1281 rabbitmq cluster in certificates domain

Exobitt:


domains/certificates/chart/values.yaml
```diff
@@ -7,3 +7,14 @@ postgres:
     name: sibedge/postgres-plv8
     tag: 15.0-3.1.4-alpine
   size: 10Gi
+rabbitmq-cluster-operator:
+  extraDeploy:
```
duizer, line 11: Do we want to define the cluster using "extraDeploy"? If possible I would prefer to have it in its own YAML-file in templates/ folder.

domains/certificates/chart/values.yaml
```diff
@@ -7,3 +7,14 @@ postgres:
...
+    - apiVersion: rabbitmq.com/v1beta1
+      kind: RabbitmqCluster
+      metadata:
+        name: certificate-rabbitmq-cluster
+      spec:
+        replicas: 1
+        rabbitmq:
+          additionalConfig: |
+            log.console.level = debug
```
duizer, line 20: Not sure that we want to have this log.console.level. I suggest that it is removed and we just use the defaults

domains/certificates/chart/values.yaml
Exobitt: Marked as resolved, as this pr is still a draft and not done yet.

domains/certificates/chart/values.yaml
Exobitt: Marked as resolved, as this pr is still a draft and not done yet.

PR #229, opened 2023-02-13
1281 add rabbitmq cluster

Exobitt:


duizer (approved):

domains/certificates/chart/templates/rabbitmq-cluster.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: rabbitmq.com/v1beta1
+kind: RabbitmqCluster
+metadata:
+  name: {{ .Release.Name }}-rabbitmq-cluster
+spec:
```
duizer, line 5: Which version of RabbitMQ is used? Can we somehow pin it?

duizer (changes requested):

domains/certificates/chart/templates/rabbitmq-cluster.yaml
duizer: It is determined by the operator. So long the operator is in the same operator version, then the cluster (or newly created clusters) will have the same rabbitmq version. 
For version 3.2.2 of https://artifacthub.io/packages/helm/bitnami/rabbitmq-cluster-operator, this will be bitnami/rabbitmq:3.10.14-debian-11-r1

duizer (approved):

anna-knudsen (approved):

PR #233, opened 2023-02-14
1304 refactor certificate domain

ckr123:


domains/certificates/Query.API/API.AppTests/CertificateIssuingTests.cs
```diff
@@ -14,17 +15,25 @@
 
 namespace API.AppTests;
 
-public sealed class CertificateIssuingTests : IClassFixture<QueryApiWebApplicationFactory>, IClassFixture<MartenDbContainer>, IDisposable
+public sealed class CertificateIssuingTests :
+    IClassFixture<QueryApiWebApplicationFactory>,
+    IClassFixture<MartenDbContainer>,
+     IClassFixture<RabbitMqContainer>,
```
duizer, line 21: Minor thing: Indentation is a bit funny here

domains/certificates/Query.API/API/RabbitMq/Configurations/RabbitMqOptions.cs
```diff
@@ -0,0 +1,11 @@
+namespace API.RabbitMq.Configurations;
+
+public class RabbitMqOptions
+{
+    public const string RabbitMq = "RabbitMq";
+
+    public string Password { get; set; } = string.Empty;
+    public string Username { get; set; } = string.Empty;
+    public string Host { get; set; } = string.Empty;
+    public string Port { get; set; } = string.Empty;
```
duizer, line 10: Port should be changed to a integer (or maybe a ushort) in stead of a string.

domains/certificates/Query.API/API/appsettings.Development.json
```diff
@@ -10,5 +10,11 @@
     },
     "Datasync": {
         "Url": "http://data-sync-api"
+    },
+    "RabbitMq": {
+        "Host": "localhost",
+        "Port": "5672",
+        "Username": "guest",
+        "Password": "guest"
```
duizer, line 18: What about our docker-compose setup? It is not part of this PR, so I guess that these values will only work if you manually start a rabbitmq instance.

domains/certificates/Query.API/API.AppTests/SwaggerTests.cs
```diff
@@ -9,11 +10,18 @@
 namespace API.AppTests;
 
 [UsesVerify]
-public class SwaggerTests : IClassFixture<QueryApiWebApplicationFactory>
+public class SwaggerTests : IClassFixture<QueryApiWebApplicationFactory>, IClassFixture<RabbitMqContainer>
```
duizer, line 13: Is the RabbitMqContainer needed in order for the swagger tests to work? If yes, is that due to the WaitUntilStarted trick we did in QueryApiWebApplicationFactory ?

domains/certificates/Query.API/API.AppTests/Infrastructure/RabbitMqContainer.cs
```diff
@@ -0,0 +1,32 @@
...
+            {
+                Username = "guest",
+                Password = "guest"
+            })
+            .Build();
+
+    public string Username => testContainer.Username;
+    public string Password => testContainer.Password;
+
+
```
duizer, line 25: Minor thing: These spaces can be deleted

domains/certificates/Query.API/API.AppTests/Infrastructure/RabbitMqContainer.cs
```diff
@@ -0,0 +1,32 @@
...
+public class RabbitMqContainer : IAsyncLifetime
+{
+    private readonly RabbitMqTestcontainer testContainer;
+
+    public RabbitMqContainer() =>
+        testContainer = new TestcontainersBuilder<RabbitMqTestcontainer>()
+            .WithMessageBroker(new RabbitMqTestcontainerConfiguration
+            {
+                Username = "guest",
+                Password = "guest"
```
duizer, line 18: Since username and password is exposed as properties, is it necessary to set them here? Will they be null otherwise?

domains/certificates/Query.API/API.AppTests/ContractTests.cs
```diff
@@ -5,22 +5,27 @@
...
+    public ContractTests(QueryApiWebApplicationFactory factory, MartenDbContainer marten, RabbitMqContainer rabbitMqContainer)
     {
         dataSyncWireMock = new DataSyncWireMock(port: 9001);
         this.factory = factory;
         this.factory.MartenConnectionString = marten.ConnectionString;
         this.factory.DataSyncUrl = dataSyncWireMock.Url;
+        this.factory.RabbitMqHost = rabbitMqContainer.Hostname;
+        this.factory.RabbitMqPort = rabbitMqContainer.Port.ToString();
+        this.factory.RabbitMqUsername = rabbitMqContainer.Username;
+        this.factory.RabbitMqPassword = rabbitMqContainer.Password;
```
duizer, line 28: Suggestion: Instead of having 4 properties to set on factory we can add a method that takes all 4 arguments. This ensures that someone does not forget to set 1 of the 4 properties.

domains/certificates/Query.API/API.AppTests/CertificateIssuingTests.cs
```diff
@@ -14,17 +15,25 @@
...
+    IDisposable
 {
     private readonly QueryApiWebApplicationFactory factory;
     private readonly DataSyncWireMock dataSyncWireMock;
 
-    public CertificateIssuingTests(QueryApiWebApplicationFactory factory, MartenDbContainer martenDbContainer)
+    public CertificateIssuingTests(QueryApiWebApplicationFactory factory, MartenDbContainer martenDbContainer, RabbitMqContainer rabbitMqContainer)
     {
-        dataSyncWireMock = new DataSyncWireMock(port: 9002);
+        dataSyncWireMock = new DataSyncWireMock(port: 9003);
```
duizer, line 29: Side note: I remember now why I set this to 9002. ContractTests uses port 9001. Since it is possible that CertificateIssuingTests and ContractTests execute in parallel, then the port should unique for the test class.
I will try to see if there is a way to both randomize the port and check if the port is open in Wiremock

domains/certificates/Query.API/API/Program.cs
```diff
@@ -57,7 +57,7 @@
 
 builder.Services.AddHealthChecks();
 
-builder.Services.AddIntegrationEventBus();
+builder.Services.AddRabbitMq(builder.Configuration);
```
duizer, line 60: Will the updated diagrams and documentation be part of this PR?

duizer (changes requested):

domains/certificates/Query.API/API.AppTests/Infrastructure/RabbitMqContainer.cs
```diff
@@ -0,0 +1,32 @@
...
+using Xunit;
+
+namespace API.IntegrationTest.Infrastructure;
+
+public class RabbitMqContainer : IAsyncLifetime
+{
+    private readonly RabbitMqTestcontainer testContainer;
+
+    public RabbitMqContainer() =>
+        testContainer = new TestcontainersBuilder<RabbitMqTestcontainer>()
```
duizer, line 14: Which docker image and version of RabbitMQ will this spin up? Is it possible to set it?
In production, the RabbitMQ version is determined by the RabbitMQ operator. For the version we are using, this will use 
this docker image+tag: "bitnami/rabbitmq:3.10.14-debian-11-r1"

domains/certificates/Query.API/API.AppTests/Infrastructure/QueryApiWebApplicationFactory.cs
```diff
@@ -23,14 +23,24 @@ public class QueryApiWebApplicationFactory : WebApplicationFactory<Program>
...
+        builder.UseSetting("RabbitMq:Password", RabbitMqPassword);
+        builder.UseSetting("RabbitMq:Username", RabbitMqUsername);
+        builder.UseSetting("RabbitMq:Host", RabbitMqHost);
+        builder.UseSetting("RabbitMq:Port", RabbitMqPort);
 
         builder.ConfigureTestServices(services =>
         {
+            //  Ensure masstransit bus is started when we run our health checks
+            services.AddOptions<MassTransitHostOptions>().Configure(options => options.WaitUntilStarted = true);
             //Remove DataSyncSyncerWorker
```
anna-knudsen, line 44: Unnecessary comment? (Remove DataSyncSyncerWorker)

doc/architecture/domains/certificates.md
```diff
@@ -17,9 +17,7 @@ The reason for this split is certain constraints. The constraints are:
 
 ## Component diagram: Certificate API
 
-The component diagram shows a first iteration which is based on an in-memory integration event bus. A consequence of using the in-memory implementation is that all components that is dependent on the integration event bus must be in same container.
-
-Components that is used for mocking and will be replaced or discarded at a later are marked with its own color in the diagram.
+The component diagram shows how the solution works, based on a RabbitMQ message broker that publishes the events received from the DataSyncSyncer.
```
duizer, line 20: In the code changes in the PR, the name "RabbitMQ" is used (domains/certificates/Query.API/API/RabbitMq/Startup.cs), but in the diagram it is called "Message Broker". Before it was called "Integration Event Bus" both in the code and in the diagram. Is it possible that we can have that the name of a component (or container) is named the same in both code and in the diagram?

duizer (changes requested):

domains/certificates/Query.API/API/Program.cs
```diff
@@ -57,7 +57,7 @@
 
 builder.Services.AddHealthChecks();
 
-builder.Services.AddIntegrationEventBus();
```
duizer, line 60: We have a shared library named "IntegrationEvents" with events we expected to receive on the Integration Event Bus. Given that we here delete the concept "Integration Event Bus" from the code, should we not also do something with event(s) under the library "IntegrationEvents"?

domains/certificates/Query.API/API.AppTests/HealthTests.cs
```diff
@@ -1,17 +1,22 @@
 using System.Net;
 using System.Threading.Tasks;
 using API.AppTests.Infrastructure;
+using API.IntegrationTest.Infrastructure;
 using Xunit;
 
 namespace API.AppTests;
 
-public class HealthTests : IClassFixture<QueryApiWebApplicationFactory>
+public class HealthTests : IClassFixture<QueryApiWebApplicationFactory>, IClassFixture<RabbitMqContainer>
```
duizer, line 9: Now that we a way of getting the application in an unhealthy state (by not having a RabbitMQ container running), should we have a test that checks that the application becomes unhealthy?

domains/certificates/Query.API/API/appsettings.Development.json
```diff
@@ -10,5 +10,11 @@
     },
     "Datasync": {
         "Url": "http://data-sync-api"
+    },
+    "RabbitMq": {
+        "Host": "localhost",
+        "Port": "5672",
```
duizer, line 16:
```suggestion
        "Port": 5672,
```

domains/certificates/Query.API/API.AppTests/ContractTests.cs
duizer: If it is a method on the factory, then that method would have to store the arguments in backing fields, which is then used in the factory's ConfigureTestServices.

A maybe better approach is to expose a single property on the factory of type RabbitMqOptions. This type has all the properties anyway.

domains/certificates/Query.API/API.AppTests/Infrastructure/RabbitMqContainer.cs
ckr123: It will spin up "rabbitmq:3.7.28" - are you suggesting being specific on the image? 
Then we will need to maintain it

domains/certificates/Query.API/API.AppTests/Infrastructure/RabbitMqContainer.cs
ckr123: Yes it is necessary, else testcontainer will look for some env vars that doesnt exists

domains/certificates/Query.API/API.AppTests/SwaggerTests.cs
ckr123: Yes else it will keep waiting

domains/certificates/Query.API/API.AppTests/HealthTests.cs
ckr123: Good point, i do have some troubles getting the health check to act as i want it to act. :) Maybe we can have a chat about it?

domains/certificates/Query.API/API.AppTests/Infrastructure/RabbitMqContainer.cs
duizer: Yes, I am suggesting that it set to the same image as will be used in production. The same also goes for the docker-compose image for RabbitMQ.
Unfortunately, it is determined by the version of RabbitMQ operator and is defined in eo-base-environment. So it will be a "magic string", but cannot see a way out of solving that right now.

domains/certificates/Query.API/API.AppTests/CertificateIssuingTests.cs
duizer: No. But removing the port number inside DataSyncWireMock should mean that it gets a random port assigned. I have not tried if there should be any issues. You are welcome to do so. If you choose not, then close the conversation.

domains/certificates/Query.API/API.AppTests/HealthTests.cs
duizer: Sure

domains/certificates/Shared/RabbitMqEvents/EnergyMeasuredIntegrationEvent.cs
```diff
@@ -1,4 +1,4 @@
-namespace IntegrationEvents;
+namespace RabbitMqEvents;
```
duizer, line 1: Have you considered other names than RabbitMqEvents? I was thinking something like "MeasurementEvents" or "MeasurementMessages" as this is releated to measurements.

domains/certificates/Query.API/API.AppTests/CertificateIssuingTests.cs
duizer: I tried just removing the port, but our tests then fails. We/I have to do a bit more investigation to why that is. But that is outside of the scope of this task. I am closing the conversation.

domains/certificates/Shared/RabbitMqEvents/EnergyMeasuredIntegrationEvent.cs
ckr123: renamed to MeasurementEvents

domains/certificates/Query.API/API.AppTests/Infrastructure/RabbitMqContainer.cs
ckr123: This will introduce some errors with M1 processors

domains/certificates/Query.API/API.AppTests/HealthTests.cs
ckr123: As agreed an unhealthtests class has been created

domains/certificates/Query.API/API.AppTests/Infrastructure/QueryApiWebApplicationFactory.cs
```diff
@@ -24,24 +25,44 @@ public class QueryApiWebApplicationFactory : WebApplicationFactory<Program>
...
+            {
+                //  Ensure masstransit bus is started when we run our health checks
+                services.AddOptions<MassTransitHostOptions>().Configure(options => options.WaitUntilStarted = true);
+            }
             //Remove DataSyncSyncerWorker
             services.Remove(services.First(s => s.ImplementationType == typeof(DataSyncSyncerWorker)));
         });
     }
 
+    private bool IsRabbitMqSet() => RabbitMqSetup != null;
```
duizer, line 57: Could we make the property nullable instead? That would also fix the code smell warning from SonarCloud.

domains/certificates/Query.API/API.AppTests/UnhealthTests.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace API.AppTests;
+
+public class UnhealthTests : IClassFixture<QueryApiWebApplicationFactory>
+{
+    private readonly QueryApiWebApplicationFactory factory;
+
+    public UnhealthTests(QueryApiWebApplicationFactory factory)
+    {
+        this.factory = factory;
```
duizer, line 14: If the `RabbitMqSetup` property is nullable, we could here have a line like `this.factory.RabbitMqSetup = null`, which indicates to the reader that there is no RabbitMQ. What do you think?

domains/certificates/Query.API/API.AppTests/SwaggerTests.cs
```diff
@@ -9,11 +11,21 @@
...
```
duizer, line 14: The code has now been changed in the factory to not wait for the bus, if RabbitMqSetup is null. So can we remove the RabbitMqContainer here from this test now?

domains/certificates/Query.API/API.AppTests/Infrastructure/RabbitMqContainer.cs
```diff
@@ -0,0 +1,33 @@
...
+    public string Username => testContainer.Username;
+    public string Password => testContainer.Password;
+
+
+    public string Hostname => testContainer.Hostname;
+    public int Port => testContainer.Port;
+
+    public async Task InitializeAsync() => await testContainer.StartAsync();
+
+    public Task DisposeAsync() => testContainer.StopAsync(); //  testContainer.DisposeAsync().AsTask();
```
duizer, line 32:
```suggestion
    public Task DisposeAsync() => testContainer.StopAsync();
```

domains/certificates/Query.API/API.AppTests/Infrastructure/RabbitMqContainer.cs
```diff
@@ -0,0 +1,33 @@
...
+                Password = "guest"
+            })
+            .Build();
+
+    public string Username => testContainer.Username;
+    public string Password => testContainer.Password;
+
+
+    public string Hostname => testContainer.Hostname;
+    public int Port => testContainer.Port;
```
duizer, line 28:
```suggestion
    public string Password => testContainer.Password;
    public string Hostname => testContainer.Hostname;
    public int Port => testContainer.Port;
```

domains/certificates/Query.API/API.AppTests/Infrastructure/QueryApiWebApplicationFactory.cs
```diff
@@ -24,24 +25,44 @@ public class QueryApiWebApplicationFactory : WebApplicationFactory<Program>
...
 
+    public RabbitMqOptions RabbitMqSetup { get; set; }
+
+
     protected override void ConfigureWebHost(IWebHostBuilder builder)
     {
         builder.UseSetting("ConnectionStrings:Marten", MartenConnectionString);
         builder.UseSetting("Datasync:Url", DataSyncUrl);
 
+        if (IsRabbitMqSet())
```
duizer, line 36: I think that as a default the ASPNETCORE_ENVIRONMENT will be set to development. If that is right, the application **may** (I am not sure) use the values listed in appsettings.development.json in QueryAPI-project. 
If not overwritten (e.g. when IsRabbitSet() is false), then will it have null/empty values or use the ones from appsettings.development.json? I am not sure. But if it uses appsettings.development.json values, then someone running the test locally will have strange test results if there is a rabbitmq running.

Doest it make sense?

domains/certificates/Query.API/API.AppTests/SwaggerTests.cs
ckr123: Seems some of the test still want to spin up rabbitmq. How is the lifecycle if these things?

domains/certificates/Query.API/API.AppTests/Infrastructure/QueryApiWebApplicationFactory.cs
ckr123: Not sure if it make sense, but no problems here when running a rabbitmq while running the tests

duizer (approved):

PR #251, opened 2023-02-28
1334 create endpoint for transfer certificate

Exobitt:


domains/certificates/Query.API/API/Query.API/Controllers/TransferCertificateController.cs
```diff
@@ -0,0 +1,50 @@
...
+[Authorize]
+[ApiController]
+public class TransferCertificateController : ControllerBase
+{
+    [HttpPost]
+    [ProducesResponseType(200)]
+    [Route("api/certificates/production/transfer")]
+    public async Task<IActionResult> TransferCertificate(
+        [FromBody] TransferCertificate transferCertificate,
+        [FromServices] ITransferCertificateService service
```
duizer, line 24: As far as I remember, it is possible to get IRequestClient from the service collection:

```suggestion
        [FromServices] IRequestClient<TransferCertificate> requestClient
```

PR #252, opened 2023-02-28
Add token refresh endpoint

MortMH:


domains/auth/Auth.API/API/Controllers/RefreshController.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.Extensions.Options;
+
+namespace API.Controllers;
+
+[Authorize]
+[ApiController]
+public class RefreshController : ControllerBase
+{
+    [HttpGet()]
+    [Route("auth/refresh")]
```
CodeReaper, line 14: See route here: https://github.com/Energinet-DataHub/energy-origin/blob/main/doc/architecture/domains/auth.md#get-apiauthtoken

domains/auth/Auth.API/API/Controllers/RefreshController.cs
```diff
@@ -0,0 +1,33 @@
...
+        var token = tokenIssuer.Issue(oldDescriptor);
+
+        accessor.HttpContext!.Response.Cookies.Append("Authentication", token, new CookieOptions
+        {
+            IsEssential = true,
+            Secure = true,
+            Expires = DateTimeOffset.UtcNow.Add(tokenOptions.Value.CookieDuration)
+        });
+
+        return NoContent();
```
CodeReaper, line 31: You are supposed to return the token (not create a cookie).

domains/auth/Auth.API/API/Controllers/RefreshController.cs
```diff
@@ -0,0 +1,33 @@
...
+    [HttpGet()]
+    [Route("auth/refresh")]
+    public ActionResult RefreshAccessToken(
+        IHttpContextAccessor accessor,
+        IOptions<TokenOptions> tokenOptions,
+        IUserDescriptMapper descriptMapper,
+        ITokenIssuer tokenIssuer)
+    {
+        var oldDescriptor = descriptMapper.Map(User) ?? throw new NullReferenceException($"UserDescriptMapper failed: {User}");
+        var token = tokenIssuer.Issue(oldDescriptor);
```
CodeReaper, line 22: This endpoint cannot simply re-issue. This endpoint handles the issue with not regressing an issued token to one where terms are not accepted. Note this will likely require looking into beginning to create/define scope claims, i.e. the ones that we will need to publish to a shared library eventually.

CodeReaper (changes requested):

domains/auth/Auth.API/API/Controllers/TokenController.cs
```diff
@@ -0,0 +1,36 @@
...
+        IUserService userService,
+        ITokenIssuer tokenIssuer)
+    {
+        var descriptor = descriptMapper.Map(User) ?? throw new NullReferenceException($"UserDescriptMapper failed: {User}");
+        var versionBypass = false;
+
+        if (descriptor.Id is not null)
+        {
+            var user = await userService.GetUserByIdAsync(descriptor.Id.Value) ?? throw new NullReferenceException($"GetUserByIdAsync() returned null: {descriptor.Id.Value}");
+            descriptor = descriptMapper.Map(user, descriptor.AccessToken!, descriptor.IdentityToken!, false);
```
CodeReaper, line 28: I would prefer to decrypt the tokens here. The boolean means you can do it incorrectly.

domains/auth/Auth.API/API/Utilities/TokenIssuer.cs
```diff
@@ -39,11 +39,16 @@ private static SigningCredentials CreateSigningCredentials(TokenOptions options)
...
+    private static UserState ResolveState(TermsOptions options, UserDescriptor descriptor, bool versionBypass)
     {
         var version = descriptor.AcceptedTermsVersion;
 
-        var scope = version == options.CurrentVersion ? "terms dashboard production meters certificates" : "terms";
+        var allScopes = "terms dashboard production meters certificates";
+
+        var scope = versionBypass
+            ? allScopes
+            : version == options.CurrentVersion ? allScopes : "terms";
```
CodeReaper, line 50: "Double inline if" - can we do anything else? :)

domains/auth/Auth.API/API/Controllers/TokenController.cs
```diff
@@ -0,0 +1,36 @@
...
+            versionBypass = true;
```
CodeReaper, line 29: Just inspecting the user id is not enough. We _must_ inspect the scope claim to figure out if we had accepted terms at the beginning of the current "session". (The id is missing only when the user is new and doing their first login - the second time they log in the id will be present)

PR #256, opened 2023-03-06
Api snapshot test

robertrosborg:


CodeReaper:
@robertrosborg It is fine to include emissions and measurements, but the task is to include the snapshot tests in sheath domain. :-)

PR #262, opened 2023-03-09
1362 update certificatesbyownerview projection based

ckr123:


domains/certificates/Query.API/API/Query.API/Projections/CertificatesTransferProjection.cs
```diff
@@ -0,0 +1,89 @@
...
+using API.Query.API.Projections.Views;
+using CertificateEvents;
+using CertificateEvents.Exceptions;
+using Marten;
+using Marten.Events;
+using Marten.Events.Projections;
+
+namespace API.Query.API.Projections;
+
+public class CertificatesTransferProjection : IProjection
```
duizer, line 14: There is two projections (CertificatesTransferProjection and CertificatesByOwnerProjection) handling/updating the same view (CertificatesByOwnerView). I would think that having a single projection handling a single view is easier to maintain and understand - also I am not sure, if there will be issues with both projections "fighting" about the same view.
Is it possible to have everything related to CertificatesByOwnerView in this projection (CertificatesTransferProjection)?

domains/certificates/Shared/CertificateEvents/ProductionCertificateTransferred.cs
```diff
@@ -6,12 +6,5 @@ namespace CertificateEvents;
...
     string Source,
-    string Target,
-
-    // Duplicated values from below
-    string GridArea,
-    Period Period,
-    Technology Technology,
-    ShieldedValue<string> ShieldedGSRN,
-    ShieldedValue<long> ShieldedQuantity
+    string Target
```
duizer, line 9: As a consequence of deleteing the duplicate values here, we should also remove the private fields in domains/certificates/Shared/CertificateEvents/Aggregates/ProductionCertificate.cs (which is also causing codesmells in Sonarcloud)

domains/certificates/Query.API/API/Query.API/Startup.cs
```diff
@@ -13,7 +13,7 @@ public static void AddQueryApi(this IServiceCollection services)
     {
         services.ConfigureMarten(o =>
         {
-            o.Projections.Add<CertificatesByOwnerProjection>(ProjectionLifecycle.Inline);
+            o.Projections.Add(new CertificatesByOwnerProjection());
```
duizer, line 16: What is the Lifecycle of the projection? It is not listed explicitly here, so is it Inline?

ckr123: It is inline

domains/certificates/Query.API/API/Query.API/Projections/CertificatesByOwnerProjection.cs
```diff
@@ -1,119 +1,153 @@
...
 
-public class CertificatesByOwnerView
-{
-    [Identity]
-    public string Owner { get; set; } = "";
+    private static CertificatesByOwnerView ApplyProductionCertificateTransferred(IDocumentOperations operations,
+        ProductionCertificateTransferred productionCertificateTransferred)
+    {
+        var source = operations.Load<CertificatesByOwnerView>(productionCertificateTransferred.Source)
+                     ?? throw new CertificateDomainException(
```
duizer, line 103: I think we should use a different exception type here as `CertificateDomainException` is thought to be used dealing with "illegal" actions on a certificate, not that a projection is failing. I suggest that we create a new type called e.g. `ProjectionException`.

domains/certificates/Query.API/API/Query.API/Projections/CertificatesByOwnerProjection.cs
```diff
@@ -1,119 +1,153 @@
...
-    public string GridArea { get; set; } = "";
-    public string TechCode { get; set; } = "";
-    public string FuelCode { get; set; } = "";
-    public CertificateStatus Status { get; set; }
-}
+    private static void RemoveCertificateFromSource(IDocumentOperations operations, CertificatesByOwnerView source,
+        ProductionCertificateTransferred productionCertificateTransferred)
+    {
+        source.Certificates.Remove(productionCertificateTransferred.CertificateId);
+        operations.Update(source);
```
duizer, line 138: It seems like "the pattern" is broken here. It appears that there should only be one update/store on line 45 with `operations.Store(view);`, but here is an extra update of a view (`source`) and is a bit hidden.

I am not sure if I have an easy fix other than going with a different approach. A different approach could be to start with an empty list of views and then let the methods for applying the events add to that list; in the end, it is just looping through that list and calling `opertions.Store(item)`

domains/certificates/Query.API/API/Query.API/Projections/CertificatesByOwnerProjection.cs
```diff
@@ -1,119 +1,153 @@
...
-        else
-        {
-            view.Certificates[@event.CertificateId] = new CertificateView
+        var events = streams
+            .SelectMany(x => x.Events)
+            .OrderBy(s => s.Sequence)
+            .Select(s => s.Data);
+
+        var view =
+            events.Aggregate<object, CertificatesByOwnerView>(null, (currentView, @event) => @event switch
```
duizer, line 32: I must admit this is a bit hard to read/understand and I think the main reason is that I do not see this as an accumulator function. Please also see my comment in the RemoveCertificateFromSource-method.

domains/certificates/Query.API/API/Query.API/Projections/CertificatesByOwnerProjection.cs
```diff
@@ -1,119 +1,153 @@
...
-        }
-        else
-        {
-            view.Certificates[@event.CertificateId] = new CertificateView
+        var events = streams
+            .SelectMany(x => x.Events)
+            .OrderBy(s => s.Sequence)
+            .Select(s => s.Data);
+
+        var view =
```
duizer, line 31: There is a hidden assumption here that we only get events for a single certificate. If there are events from more than one certificate, then this will (probably) fail. I suggest that we make this hidden assumption clear by throwing an exception in case of more than one unique certificate id

domains/certificates/Query.API/API/Query.API/Projections/CertificatesByOwnerProjection.cs
```diff
@@ -1,119 +1,186 @@
...
+        views.Add(source);
+    }
 
-public enum CertificateStatus
-{
-    Creating = 1,
-    Issued = 2,
-    Rejected = 3
-};
+    private static CertificatesByOwnerView GetCertificatesByOwnerView(IDocumentOperations operations,
```
duizer, line 174: According to SonarCloud there are no tests that cover this method. Why is that the case?

domains/certificates/Query.API/API/Query.API/Projections/CertificatesByOwnerProjection.cs
```diff
@@ -1,119 +1,186 @@
...
-using Marten.Schema;
+using Marten.Linq.LastModified;
 
 namespace API.Query.API.Projections;
 
-public class CertificatesByOwnerProjection : MultiStreamAggregation<CertificatesByOwnerView, string>
+public class CertificatesByOwnerProjection : IProjection
 {
-    public CertificatesByOwnerProjection()
+    private static readonly List<CertificatesByOwnerView> views = new();
```
duizer, line 19: By marking `views` as static, it means that it is bound to the class and not an instance of the class. I do not think that it should be static here.

domains/certificates/Query.API/API/Query.API/Projections/CertificatesByOwnerProjection.cs
```diff
@@ -1,119 +1,186 @@
...
+    private static void ApplyProductionCertificateRejected(IDocumentOperations operations,
+        ProductionCertificateRejected productionCertificateRejected)
+    {
+        var view = views
+                       .FirstOrDefault(it =>
+                           it.Certificates.ContainsKey(productionCertificateRejected.CertificateId)) ??
+                   GetCertificatesByOwnerView(operations, productionCertificateRejected.CertificateId);
+
+        view.Certificates[productionCertificateRejected.CertificateId].Status = CertificateStatus.Rejected;
+        views.Add(view);
```
duizer, line 78: If the item is already in `views`, that item will be added again to `views`. I think it would be safer to *not* have the same item multiple times in `views`

domains/certificates/Query.API/API/Query.API/Projections/CertificatesByOwnerProjection.cs
```diff
@@ -1,119 +1,186 @@
...
+    private static void ApplyProductionCertificateIssued(IDocumentOperations operations,
+        ProductionCertificateIssued productionCertificateIssued)
+    {
+        var view = views
+                       .FirstOrDefault(it =>
+                           it.Certificates.ContainsKey(productionCertificateIssued.CertificateId)) ??
+                   GetCertificatesByOwnerView(operations, productionCertificateIssued.CertificateId);
+
+        view.Certificates[productionCertificateIssued.CertificateId].Status = CertificateStatus.Issued;
+        views.Add(view);
```
duizer, line 90: Please see the comment for ApplyProductionCertificateRejected about adding to `views`

domains/certificates/Query.API/API/Query.API/Projections/CertificatesByOwnerProjection.cs
```diff
@@ -1,119 +1,186 @@
...
-{
-    [Identity]
-    public string Owner { get; set; } = "";
+    private static void ApplyProductionCertificateTransferred(IDocumentOperations operations,
+        ProductionCertificateTransferred productionCertificateTransferred)
+    {
+        var source = operations.Load<CertificatesByOwnerView>(productionCertificateTransferred.Source)
+                     ?? throw new ProjectionException(
+                         productionCertificateTransferred.CertificateId,
+                         $"Cannot transfer from {productionCertificateTransferred.Source}. {productionCertificateTransferred.CertificateId} cannot be found"
```
duizer, line 135:
```suggestion
                         $"Cannot transfer from {productionCertificateTransferred.Source}. View for {productionCertificateTransferred.Source} cannot be found"
```

domains/certificates/Query.API/API/Query.API/Projections/CertificatesByOwnerProjection.cs
```diff
@@ -1,119 +1,186 @@
...
+    }
 
-public class CertificatesByOwnerView
-{
-    [Identity]
-    public string Owner { get; set; } = "";
+    private static void ApplyProductionCertificateTransferred(IDocumentOperations operations,
+        ProductionCertificateTransferred productionCertificateTransferred)
+    {
+        var source = operations.Load<CertificatesByOwnerView>(productionCertificateTransferred.Source)
```
duizer, line 132: There is no attempt to get the view from `views` here. That works for now as the the Transferred event will be the only event in the transaction. But it is different compared to how Issued and Rejected events are applied. Would it make sense to have the same logic wrt. `views` for the Apply-methods for Issued, Rejected and Transferred?

domains/certificates/Query.API/API.AppTests/CertificatesByOwnerProjectionTest.cs
```diff
@@ -0,0 +1,112 @@
...
+    public CertificatesByOwnerProjectionTest(MartenDbContainer martenDbContainer)
+    {
+        using var store = DocumentStore.For(options =>
+        {
+            options.Connection(martenDbContainer.ConnectionString);
+            options.Projections.Add(new CertificatesByOwnerProjection());
+        });
+
+        repository = new ProductionCertificateRepository(store);
+        session = store.LightweightSession();
```
duizer, line 31: If the session implements IDisposable, it should be disposed.

domains/certificates/Query.API/API.AppTests/CertificatesByOwnerProjectionTest.cs
```diff
@@ -0,0 +1,112 @@
...
+        await repository.Save(productionCertificate);
+
+        var certificate = await repository.Get(productionCertificate.Id);
+        certificate!.Transfer(source, target);
+        await repository.Save(certificate);
+
+        var targetView = session.Load<CertificatesByOwnerView>(target);
+        var sourceView = session.Load<CertificatesByOwnerView>(source);
+
+        sourceView?.Certificates.Should().BeEmpty();
```
duizer, line 51:
```suggestion
        sourceView!.Certificates.Should().BeEmpty();
```

duizer (changes requested):

domains/certificates/Query.API/API/Query.API/Projections/Exceptions/ProjectionException.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using System.Runtime.Serialization;
+
+namespace API.Query.API.Projections.Exceptions;
+
+[Serializable]
+public class ProjectionException : Exception
```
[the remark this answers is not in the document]
ckr123: Something went wrong when running sonarcloud. a rerun fixed the problem

duizer (approved):

duizer (approved):

PR #274, opened 2023-03-24
Document issuer key generation

duizer:


Exobitt (approved):

domains/certificates/README.md
```diff
@@ -1,6 +1,18 @@
...
+
+An Ed25519 key must be generated for the issuing body of certificates, which in this case is Energinet. Project Origin Registry requires that the signature algorithm is Ed25519. To be consistent with the Auth domain, the format used is PEM. The following generates a key and does base64 encoding:
+
+```
+openssl genpkey -algorithm ed25519 | base64 -w 0
+```
+
+A key is generated and added as a sealed secret in eo-base-environment. This is the private key. At this point in time, the same key is used for all Energy Origin environments.
+
+Project Origin Registry must know the public key for the issuer for the relevant grid areas (e.g. "DK1" and "DK2") and the format here must be "RawPublicKey" (see https://nsec.rocks/docs/api/nsec.cryptography.keyblobformat).
```
ckr123, line 14: Probally not the right place to have this question, but why are PO using the "not recommended" RawPublicKey format?

ckr123 (approved):

PR #275, opened 2023-03-24
Add state and overrides

CodeReaper:


domains/auth/Auth.API/Unit.Tests/Controllers/LoginControllerTests.cs
```diff
@@ -48,10 +51,45 @@ public async Task LoginAsync_ShouldReturnRedirectToAuthority_WhenInvoked()
...
+        var query = HttpUtility.UrlDecode(uri.Query);
+        var map = QueryHelpers.ParseNullableQuery(query);
+
+        Assert.NotNull(map);
+        Assert.True(map.ContainsKey("state"));
+
+        var decoded = OidcState.Decode(map["state"]);
+        Assert.NotNull(decoded);
+        Assert.Equal(state, decoded.State);
+        Assert.Equal(redirectionUri, decoded.RedirectionUri);
```
CodeReaper, line 90: The state in the query string should be equal to original state. OidcState is only for the "inside".

domains/auth/Auth.API/Unit.Tests/Controllers/OidcControllerTests.cs
```diff
@@ -96,14 +94,98 @@ public async Task CallbackAsync_ShouldReturnRedirectToFrontendWithCookie_WhenInv
...
+        var redirectionUri = "https://example.com";
+        var oidcState = new OidcState(State: null, RedirectionUri: redirectionUri);
+
+        var action = await new OidcController().CallbackAsync(cache, factory, mapper, service, issuer, oidcOptions, logger, Guid.NewGuid().ToString(), null, null, oidcState.Encode());
+
+        Assert.NotNull(action);
+        var result = (RedirectResult)action;
+
+        var uri = new Uri(result.Url);
+        Assert.Equal(oidcOptions.Value.FrontendRedirectUri.Host, uri.Host);
```
CodeReaper, line 143: This should be equal to the given redirection uri.

Frydensberg1 (approved):

PR #278, opened 2023-03-27
1400 - Key issuer for Certificates

Exobitt:


Exobitt:
This is not nearly done. Currently, i saved it inside the contracts service. But that's not right, as more services might use it. So im planning on moving it into its own directory.

domains/certificates/Query.API/API/KeyIssuer/KeyIssuer.cs
```diff
@@ -0,0 +1,71 @@
...
+        var data = Encoding.UTF8.GetBytes(meteringPointOwner);
+        var signature = algorithm.Sign(key, data);
+
+        var encodedPublicKey = Encode(key.PublicKey.Export(KeyBlobFormat.NSecPublicKey));
+
+        var keyDocument = new KeyIssuingDocument
+        {
+            MeteringPointOwner = meteringPointOwner,
+            PublicKey = encodedPublicKey,
+            Signature = Encode(signature)
```
duizer, line 40: The private key is not stored here. It should be stored, otherwise it is lost

domains/certificates/Query.API/API/KeyIssuer/Startup.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace API.KeyIssuer;
+
+public static class Startup
+{
+    public static void AddKeyIssuerService(this IServiceCollection services) =>
+        services.ConfigureMarten(o =>
+        {
+            o.Schema
+                .For<KeyIssuingDocument>()
+                .UniqueIndex(UniqueIndexType.Computed, "uidx_meteringpointowner");
```
duizer, line 14: There is no identity defined for KeyIssuingDocument, but it should most likely be "MeteringPointOwner". There are different ways of setting the identity if not naming a property "Id": https://martendb.io/documents/identity.html#overriding-the-choice-of-id-property-field

ckr123:
> 

Is this still relevant?

ckr123:
I think it would be nice with some nice test to cover this

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncSyncerWorker.cs
```diff
@@ -39,6 +45,20 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
             var allContracts = await GetAllContracts(stoppingToken);
             foreach (var contract in allContracts)
             {
+                if (string.IsNullOrEmpty(contract.PublicKey))
+                {
+                    using var scope = scopeFactory.CreateScope();
+                    var keyIssuer = scope.ServiceProvider.GetService<IKeyIssuer>();
+                    var certificateIssuingContractRepository = scope.ServiceProvider.GetService<ICertificateIssuingContractRepository>();
```
ckr123, line 52: could this be moved out in a separate service that is run on startup and  then we can delete it afterwards?

domains/certificates/Query.API/API/KeyIssuer/KeyHandler.cs
```diff
@@ -0,0 +1,84 @@
...
+    private readonly SignatureAlgorithm algorithm;
+    private readonly IKeyIssuingRepository repository;
+
+    public KeyHandler(IKeyIssuingRepository repository)
+    {
+        algorithm = SignatureAlgorithm.Ed25519;
+        this.repository = repository;
+    }
+
+    public async Task<string> Create(string meteringPointOwner, CancellationToken cancellationToken)
```
ckr123, line 22: It seems to me that this method has several responsibilities.
I would suggest having a method to create the key, save it in the documentstore and one to fetch the publickey

domains/certificates/Query.API/API/KeyIssuer/KeyHandler.cs
```diff
@@ -0,0 +1,84 @@
...
+        var decodedPublicKey = PublicKey.Import(
+            algorithm,
+            Decode(publicKey),
+            KeyBlobFormat.NSecPublicKey
+            );
+
+        var keyDocument = await repository.GetByMeteringPointOwner(meteringPointOwner, cancellationToken);
+        if (keyDocument == null)
+        {
+            throw new Exception("Not existing");
```
ckr123, line 72: This result in a codesmell i think we should handle

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncSyncerWorker.cs
ckr123: It was actual the whole code blob as this is per my understanding something that only has to be run once for every existing contract.
I dno if it will fix the scope issue

domains/certificates/Query.API/API/KeyIssuer/KeyHandler.cs
ckr123: I think it is a clear violation of the SOLID principle and I believe we should strive after using that principle.
This will result in some small methods, but it will result in clean code that is easy to maintain

ckr123: I think it is kinda "hidden" that the Create method for a keyIssuer is returning the public key

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncSyncerWorker.cs
tnickelsen: Can't this be fixed with DI like any other service?

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncSyncerWorker.cs
```diff
@@ -39,6 +45,20 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
             var allContracts = await GetAllContracts(stoppingToken);
             foreach (var contract in allContracts)
             {
+                if (string.IsNullOrEmpty(contract.PublicKey))
```
tnickelsen, line 48: Have you considered going DDD on CertificateIssuingContract (contract) and other domain objects?

duizer:
From the workshop with Project Origin and Elia we learned that this is not the right principle for the private/public key. Our understanding that there needs to be private/public key per legal entity (company or a human) is incorrect. It needs to be to a private/public key per slice in a certificate. 
Martin Schmidt proposed a solution where we can calculate from a seed + extra information. This would mean that we do not need to store the keys, but can calculate them when needed.

But this leaves this us in a situation where we should re-think what we do. This PR and the task should be closed/abandoned

PR #280, opened 2023-03-28
Add registry connector

duizer:
Please also see https://github.com/Energinet-DataHub/eo-base-environment/pull/458


domains/certificates/chart/values.yaml
```diff
@@ -1,6 +1,9 @@
 api:
   replicaCount: 1
   image: {}
+registryconnector:
+  replicaCount: 3
```
ckr123, line 5: is 3 the "standard"?

domains/certificates/docker-test-env/docker-compose.yml
```diff
@@ -16,6 +16,17 @@ services:
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: postgres
 
+  registry:
+    image: ghcr.io/project-origin/electricity-server:0.1.0-alpha.18
+    ports:
+      - 8765:80
+    environment:
+      IMMUTABLELOG__TYPE: "log"
+      REGISTRIES__RegistryA__VERIFIABLEEVENTSTORE__EVENTSTORE__TYPE: "inMemory"
```
ckr123, line 25: Should all thse env vars not be upper case?

domains/certificates/chart/values.yaml
duizer: It seems to be the number used in other domains (e.g. Emissions and Measurements). I just want a number larger than 1.

domains/certificates/docker-test-env/docker-compose.yml
duizer: It is not a rule, but a convention. I choose to write it as "RegistryA" over "REGISTRYA" to make it easier to read.

domains/certificates/RegistryConnector/Worker/Worker.cs
```diff
@@ -0,0 +1,90 @@
...
+        => logger.LogInformation("Received event. Id={id}, State={state}, Error={error}", ToHex(cse.Id), cse.State, cse.Error);
+
+    private async Task SendCommand(Key ownerKey)
+    {
+        const long gsrn = 57000001234567;
+        var quantity = new ShieldedValue(150);
+
+        var commandBuilder = new ElectricityCommandBuilder();
+
+        var federatedCertifcateId = new FederatedCertifcateId(
```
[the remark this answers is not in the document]
duizer: Fixed in e20fb6c7e254cc0118084a4b175d9232dfd96499

domains/certificates/docker-test-env/docker-compose.yml
duizer: I have also tested to see if the registry is case sensitive wrt. the registry name. And it is case sensitive in the current version. So "RegistryA" is not the same as "REGISTRYA" or "registryA" or "registrya".
Since I would like for the final name to be "RegistryA", then I would like to keep this

ckr123 (approved):

PR #286, opened 2023-04-11
log running test

duizer:


domains/certificates/Query.API/API.AppTests/ContractTests.cs
```diff
@@ -61,11 +71,11 @@ public async Task CreateContract_GsrnAlreadyExistsInDb_Conflict()
 
         var body = new { gsrn, startDate = DateTimeOffset.Now.ToUnixTimeSeconds() };
 
-        var response = await client.PostAsJsonAsync("api/certificates/contracts", body);
+        using var response = await client.PostAsJsonAsync("api/certificates/contracts", body);
         response.StatusCode.Should().Be(HttpStatusCode.Created);
 
-        response = await client.PostAsJsonAsync("api/certificates/contracts", body);
-        response.StatusCode.Should().Be(HttpStatusCode.Conflict);
+        using var response2 = await client.PostAsJsonAsync("api/certificates/contracts", body);
```
ckr123, line 77: Could this be renamed to something like `conflictResponse` ? just from reading this text it is hard to figure out what response2 is compared to response

domains/certificates/Query.API/API.AppTests/Extensions/HttpClientExtensions.cs
```diff
@@ -21,12 +21,14 @@ public static async Task<T> RepeatedlyGetUntil<T>(this HttpClient client, string
...
-            if (response.StatusCode == HttpStatusCode.OK)
+            using (var response = await client.GetAsync(requestUri))
             {
-                var content = await response.Content.ReadFromJsonAsync<T>();
-                if (content != null && condition(content))
-                    return content;
+                if (response.StatusCode == HttpStatusCode.OK)
+                {
+                    var content = await response.Content.ReadFromJsonAsync<T>();
+                    if (content != null && condition(content))
```
ckr123, line 29: Could this condition be separated in a method describing what it does?

domains/certificates/Query.API/API.AppTests/Testcontainers/RabbitMqContainer.cs
```diff
@@ -0,0 +1,41 @@
...
+using Testcontainers.RabbitMq;
+using Xunit;
+
+namespace API.AppTests.Testcontainers;
+
+public partial class RabbitMqContainer : IAsyncLifetime
+{
+    private readonly global::Testcontainers.RabbitMq.RabbitMqContainer testContainer;
+
+    [GeneratedRegex("\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}:(\\d+)", RegexOptions.CultureInvariant, matchTimeoutMilliseconds: 1000)]
```
ckr123, line 13: Is it possible to put this regex in a var so it is more readable?

domains/certificates/Query.API/API.AppTests/Testcontainers/RabbitMqContainer.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Threading.Tasks;
+using API.RabbitMq.Configurations;
+using Testcontainers.RabbitMq;
+using Xunit;
+
+namespace API.AppTests.Testcontainers;
+
+public partial class RabbitMqContainer : IAsyncLifetime
+{
+    private readonly global::Testcontainers.RabbitMq.RabbitMqContainer testContainer;
```
ckr123, line 11: Why is this long type needed?

duizer: We have a class in our namespace named RabbitMqContainer, which is using a class named RabbitMqContainer in another namespace.

domains/certificates/Query.API/API.AppTests/Testcontainers/RabbitMqContainer.cs
duizer: See changes in 296d402e11ef17cbe387d86aea306a84756eca60 
The regex pattern is given in an attribute and I cannot use a variable here.

domains/certificates/Query.API/API.AppTests/ContractTests.cs
duizer: Fixed in 640a5f61d68372a04019367232d71da60ab16e4d

domains/certificates/Query.API/API.AppTests/Extensions/HttpClientExtensions.cs
duizer: `condition` is a Func delegate and this is could be named to something else. Do you have a suggestion?

domains/certificates/Query.API/API.AppTests/Testcontainers/RabbitMqContainer.cs
duizer: Side note: SonarCloud reported a code smell for not using the GeneratedRegex-attribute. Before that I created the Regex in the `Options` get-method and had the option to extract the pattern into a variable; but that resulted in a code smell.

domains/certificates/Query.API/API.AppTests/Extensions/HttpClientExtensions.cs
ckr123: No sorry as i don't know what it actually checks for.

duizer: The extention method is called `RepeatedlyGetUntil` and it will repeatedly do a GET request until the condition given in the parameter `condition` is met/true.  Please note that the condition is supplied outside of the method, so what it actually does is defined elsewhere.

ckr123 (approved):

PR #293, opened 2023-04-17
Rename to integration tests

duizer:
Also allow parallel execution of integration tests when running locally


domains/certificates/Query.API/API.IntegrationTests/API.IntegrationTests.csproj
```diff
@@ -8,6 +8,10 @@
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
+  <PropertyGroup Condition="$(CI) == true">
```
ckr123, line 11: Where is this set?

ckr123 (approved):

domains/certificates/Query.API/API.IntegrationTests/API.IntegrationTests.csproj
duizer: It is an environment variable and is a default environment variable when running a Github Actions workflow: https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

PR #303, opened 2023-04-26
Use ValidationProblemDetails as response type for 400 BadRequest

duizer:
Added Serilog.Enrichers.Span, which ensures that we have traceid and spanid in our logs. Traceid is returned as part of the body for ValidationProblemDetails

I have introduced a Swashbuckle OpertionFilter ResponseSchemaForBadRequestFilter. This automatically sets the response type for a bad request to be ValidationProblemDetails. Please let me know if this convention is preferred over having to explicitly set the response type in the [ProducesResponseType]-attribute


domains/certificates/Query.API/API/API.csproj
```diff
@@ -18,6 +18,7 @@
     <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="7.0.4" />
     <PackageReference Include="Microsoft.Extensions.Http.Polly" Version="7.0.4" />
     <PackageReference Include="Serilog.AspNetCore" Version="6.0.1" />
+    <PackageReference Include="Serilog.Enrichers.Span" Version="3.1.0" />
```
tnickelsen, line 21: Why serilog changes?

tnickelsen (changes requested):

domains/certificates/Query.API/API/API.csproj
duizer: I added Serilog.Enrichers.Span, which ensures that we have traceid and spanid in our logs. Traceid is returned as part of the body for ValidationProblemDetails. So now if someone calling our API gets a 400 Bad Request, we can find the corresponding logs from the traceid.

Exobitt (approved):

tnickelsen (approved):

PR #308, opened 2023-05-02
Improve serilog

CodeReaper:


domains/auth/Auth.API/API/Program.cs
```diff
@@ -19,16 +19,22 @@
...
 using Serilog.Formatting.Json;
 
-var logger = new LoggerConfiguration()
-    .WriteTo.Console(new JsonFormatter())
-    .CreateLogger();
-
 var builder = WebApplication.CreateBuilder(args);
 
+var log = new LoggerConfiguration()
+    .Filter.ByExcluding("RequestPath like '/health%'")
```
duizer, line 28: Should this be healtz?

duizer (dismissed):

domains/auth/Auth.API/API/Program.cs
CodeReaper: This should match `healthz`, `healthx`, `health`, `healthiness` etc.

Frydensberg1 (approved):

PR #309, opened 2023-05-03
Update editor config

CodingDegen:
While using a editor we found it had some significantly different default values for certain settings compared to Visual Studio and Visual Studio Code.

We have some settings that preserve user decided line breaks, but these are undercut by a very low setting for maximum line length, therefore we are setting this to a very large value (since it cannot be disabled).

We are setting `csharp_braces_for_ifelse` to enforce a rule that was supposed to have been set when the editorconfig was introduced. The setting requires "if" and "else" clauses to have curly brackets.


.editorconfig
```diff
@@ -82,6 +83,7 @@ csharp_prefer_simple_using_statement = true:warning
 csharp_style_prefer_switch_expression = true:suggestion
 dotnet_style_readonly_field = true:suggestion
 dotnet_style_require_accessibility_modifiers = for_non_interface_members:suggestion
+csharp_braces_for_ifelse = required
```
duizer, line 86: I am too lazy to check what each statement does in the editorconfig. Could you add a description in the PR about the changes you are suggesting?

.editorconfig
```diff
@@ -120,6 +122,8 @@ csharp_style_conditional_delegate_call = true:suggestion
 csharp_style_prefer_index_operator = false:none
 csharp_style_prefer_range_operator = false:none
 csharp_style_pattern_local_over_anonymous_function = false:none
+dotnet_diagnostic.IDE0047.severity = warning
+dotnet_diagnostic.IDE0048.severity = warning
```
CodeReaper, line 126:
```suggestion
```

These warnings are for parentheses in math expressions like: `a + (b * c)`

CodingDegen:
@cla-bot check

CodeReaper:
We are not setting `csharp_braces_for_ifelse` for now to get this PR merged faster.

CodeReaper (approved):

PR #316, opened 2023-05-11
Document decisions for transfer agreements

duizer:
This is documenting the main decisions that we have taken specifically for transfer agreements. It could be written as ADR's, but the decisions here does not really impact the overall architecture. We do not have a concept of "Domain Decision Records", but we could introduce it here and use the same template as for the ADR's.


sahma19 (approved):
approved

PR #329, opened 2023-05-30
Preview/1473 create transfer agreement

sahma19:
Added migrations, npgsql db context, services, controller with create endpoint, as well as unit, and integration test.


domains/transfer-agreements/TransferAgreements.API/API.UnitTests/Data/TransferAgreementServiceTests.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        [Fact]
+        public async Task CreateTransferAgreement_ShouldAddTransferAgreementToDatabase()
+        {
+
+            var options = new DbContextOptionsBuilder<ApplicationDbContext>()
+                .UseInMemoryDatabase(databaseName: "TestDatabase")
+                .Options;
+
+            using (var context = new ApplicationDbContext(options))
```
ckr123, line 20: You can inline this using statement.
I am not sure if this is what we want, this is spinning a memory db up. We are doing the same thing in the integration test.
Is it possible to simple mock the DB?

PR #337, opened 2023-06-01
1465 transfer agreements list

sahma19:


domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
```diff
@@ -12,23 +13,36 @@ namespace API.Controllers;
...
 {
-    private readonly ITransferAgreementService transferAgreementService;
+    private readonly ITransferAgreementRepository transferAgreementRepository;
 
-    public TransferAgreementsController(ITransferAgreementService transferAgreementService) => this.transferAgreementService = transferAgreementService;
+    public TransferAgreementsController(ITransferAgreementRepository transferAgreementRepository) => this.transferAgreementRepository = transferAgreementRepository;
 
     [HttpPost("api/transfer-agreements")]
     public async Task<ActionResult> Create([FromBody] CreateTransferAgreement request)
+
```
duizer, line 22: Please delete this empty line

domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
```diff
@@ -12,24 +13,27 @@ namespace API.Controllers;
 [ApiController]
 public class TransferAgreementsController : ControllerBase
 {
-    private readonly ITransferAgreementService transferAgreementService;
+    private readonly ITransferAgreementRepository transferAgreementRepository;
 
-    public TransferAgreementsController(ITransferAgreementService transferAgreementService) => this.transferAgreementService = transferAgreementService;
+    public TransferAgreementsController(ITransferAgreementRepository transferAgreementRepository) => this.transferAgreementRepository = transferAgreementRepository;
 
     [HttpPost("api/transfer-agreements")]
```
duizer, line 20: Please add `[ProducesResponseType(201)]`.
The open api definition says that it is returning 200 OK. Adding the attribute should set that it only returns 201 Created

domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
```diff
@@ -12,24 +13,27 @@ namespace API.Controllers;
...
+            ActorId = actor,
+            SenderId = Guid.Parse(subject),
             ReceiverTin = request.ReceiverTin
         };
 
-        var result = await transferAgreementService.CreateTransferAgreement(transferAgreement);
+        var result = await transferAgreementRepository.CreateTransferAgreement(transferAgreement);
 
-        return Ok(result);
+        return Created($"api/transfer-agreements/{result.Id}", result);
```
duizer, line 37: An object of type `TransferAgreement` is returned, which is a domain object/entity/internal data model (I am not sure what you call it). Should it not be mapped to an API model?

duizer, line 37: You are hardcoding the route, which should be fixed now or at a later point. If you had the GET endpoint, then you could use `CreatedAtRoute` or `CreatedAtAction`.

domains/transfer-agreements/TransferAgreements.API/API/Data/TransferAgreementRepository.cs
```diff
@@ -2,11 +2,11 @@
...
 
-public class TransferAgreementService : ITransferAgreementService
+public class TransferAgreementRepository : ITransferAgreementRepository
 {
     private readonly ApplicationDbContext context;
 
-    public TransferAgreementService(ApplicationDbContext context) => this.context = context;
+    public TransferAgreementRepository(ApplicationDbContext context) => this.context = context;
 
     public async Task<TransferAgreement> CreateTransferAgreement(TransferAgreement transferAgreement)
```
duizer, line 11: This method does not Create a TransferAgreement; it saves a transfer agreement or it inserts it into the repository. I would suggest renaming the method and also change the return type to `Task` instead of `Task<TransferAgreement>`

domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
```diff
@@ -12,24 +13,27 @@ namespace API.Controllers;
...
         {
-            SenderId = Guid.Parse(User.FindFirstValue("sub")),
-            ActorId = User.FindFirstValue("atr"),
-            StartDate = request.StartDate.UtcDateTime,
-            EndDate = request.EndDate.UtcDateTime,
+            StartDate = DateTimeOffset.FromUnixTimeSeconds(request.StartDate),
+            EndDate = DateTimeOffset.FromUnixTimeSeconds(request.EndDate),
+            ActorId = actor,
+            SenderId = Guid.Parse(subject),
             ReceiverTin = request.ReceiverTin
```
duizer, line 32: Is there a reason why there is no validation of the ReceiverTin?

domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
```diff
@@ -12,24 +13,27 @@ namespace API.Controllers;
...
+    private readonly ITransferAgreementRepository transferAgreementRepository;
 
-    public TransferAgreementsController(ITransferAgreementService transferAgreementService) => this.transferAgreementService = transferAgreementService;
+    public TransferAgreementsController(ITransferAgreementRepository transferAgreementRepository) => this.transferAgreementRepository = transferAgreementRepository;
 
     [HttpPost("api/transfer-agreements")]
     public async Task<ActionResult> Create([FromBody] CreateTransferAgreement request)
     {
+        var actor = User.FindActorClaim();
+        var subject = User.FindSubjectClaim();
```
duizer, line 24: As far as I understand, it should only be possible to create transfer agreements for a company and not for a private person. Will this validation be added later?

domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
sahma19: It could. But the internal model also acts as the layout of the response. One could argue that creating a separate model, which represents the resonse (Something like "CreateTransferAgreementResponse"), would establish separation of concerns, more "correctly". I am leaving it as is for now, as i will be creating new response DTO's, when implementing show and list.

sahma19: Noted. It will be fixed, while implementing the GET endpoint in #1465

domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
sahma19: Because it isn't specified, in the issue, as a subtask. Implementing validations, should in my opinion be specified in a separate issue, due to the complexity, and additional tasks, it would add to the existing issue, regarding #1473, but also #1465.

I have made a list of validations that are potentially needed, as of this moment:

**1.** A TransferAgreement between the SAME Sender, and Receiver, cannot be created if the range between StartDate and EndDate overlap, with a previously created TransferAgreements start and enddate. For example you are not able to create another transferagreement if there is still a transfer agreement that is valid, since it hasn't expired yet. The start date of the new transferagreement must be set after the enddate of the previously created transferagreement.

**2.** A TransferAgreement cannot have the same subject as sender and receiver. Which means that the subject currently logged in can't send himself transfer agreements. The UUid for the currently logged in user, is collected from the user session jwt. it should be compared to the receiverTin's associated subjectId.

**3.** A transferAgreement cannot have a start date prior to the current UTC DatetimeOffset timestamp.

**4.** A transferagreement Cannot have an enddate that is prior or the same date as the startdate. The enddate must be a later date than the startdate.

**5.** A transferagreement MUST have a startdate, enddate?, ReceiverTin,

**6.** The ReceiverTin which the user sends in the creation Dto, must be 8 digits.


The amount of validations, combined with the complexity of implementing them, means that they should be refined, and specified in a separate issue.

domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
sahma19: I cannot say, and I would be hesitant to take those kinds of decisions, in a pull request. 

If your question is whether the 2 entities represent persons or companies, then:

 **Subject** represents the _company_
 
 **Actor** represents the _person/user,_ acting in that _companys_ name.

domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
duizer: I have unresolved this as your open api definition still says it is returning 200 OK, but is actually returning 201 Created

duizer (approved):

PR #338, opened 2023-06-01
Pin dotnet version

CodeReaper:


global.json
```diff
@@ -0,0 +1,5 @@
+{
+    "sdk": {
+        "version": "7.0.201"
```
MadsThy, line 3: I see they advertise 7.0.203 as the latest 7.0.2xx security update on the official page :) https://dotnet.microsoft.com/en-us/download/dotnet/7.0

Is there a reason for sticking to 201? :)

domains/transfer-agreements/TransferAgreements.API/API/Dockerfile
```diff
@@ -18,6 +17,10 @@ FROM build AS publish
 RUN dotnet publish "API.csproj" -c Release -o /app/publish /p:UseAppHost=false
 
 FROM base AS final
+
+RUN apk add --no-cache tzdata
```
MadsThy, line 21: Is there a timezone issue? :)

domains/certificates/RegistryConnector/Worker/Dockerfile
```diff
@@ -18,6 +17,10 @@ FROM build AS publish
 RUN dotnet publish "Worker.csproj" -c Release -o /app/publish /p:UseAppHost=false
 
 FROM base AS final
+
+RUN apk add --no-cache tzdata
```
MadsThy, line 21: Do we have a timezone issue for logs, or? :)

.github/workflows/build-domain.yaml
```diff
@@ -126,30 +127,43 @@ jobs:
...
+          dotnet=${{ matrix.build.dotnetVersion }}
+
+          {
+            echo "service=${service}"
+            echo "project=${PROJECT}"
+            echo "docker=${docker}"
+            echo "domain=${domain}"
+            echo "tag=${tag}"
+            echo "name=${name}"
+            echo "dotnet=${dotnet::${#dotnet}-2}"
```
MadsThy, line 142: whats the -2 for? :)

MadsThy (dismissed):
Just had a few questions for clarification :)

global.json
CodeReaper: Yes, it seems the latest devcontainer available is still at 201. The SDK version is only used during builds and the runtime version used is the latest one.

domains/transfer-agreements/TransferAgreements.API/API/Dockerfile
CodeReaper: There has been issues with time zone data before when switching to alpine. It is added to be safe.

domains/certificates/RegistryConnector/Worker/Dockerfile
CodeReaper: There has been issues with time zone data before when switching to alpine. It is added to be safe.

.github/workflows/build-domain.yaml
CodeReaper: Almost, it turns `7.0.201` into `7.0.2`. The first one being the SDK version and the later one being the runtime version.

.github/workflows/build-domain.yaml
```diff
@@ -126,30 +127,43 @@ jobs:
...
+          } >> $GITHUB_OUTPUT
+
+      - name: Pin dotnet version
+        shell: bash
+        working-directory: ${{ steps.resolver.outputs.service }}
+        run: echo '{"sdk":{"version":"${{ inputs.dotnet-version }}"}}' > global.json
```
duizer, line 148: I do not understand why it is needed to generate global.json files at all? Why not just utilize how the dotnet tooling works where it finds the nearest global.json?

CodeReaper: There needs to be a global.json file "locally" for it to be copied during docker builds. It is only needed if we want to ensure those are built using the specified version.

.github/workflows/build-libraries.yaml
```diff
@@ -60,18 +60,29 @@ jobs:
...
 
       - name: Resolve version
         id: resolve-version
         shell: bash
         run: echo "result=$(yq '.version' ${{ env.SOLUTION }}/version.yaml)" >> $GITHUB_OUTPUT
 
+      - name: Pin dotnet version
+        shell: bash
+        working-directory: ${{ env.SOLUTION }}
+        run: echo '{"sdk":{"version":"${{ env.SDK_VERSION }}"}}' > global.json
```
duizer, line 79: Should "rollFoward" be set to "disable" here?

.github/workflows/build-domain.yaml
```diff
@@ -126,30 +128,43 @@ jobs:
...
+            echo "domain=${domain}"
+            echo "tag=${tag}"
+            echo "name=${name}"
+            echo "dotnet=${dotnet::${#dotnet}-2}"
+          } >> $GITHUB_OUTPUT
+
+      - name: Pin dotnet version
+        shell: bash
+        working-directory: ${{ steps.resolver.outputs.service }}
+        run: echo '{"sdk":{"rollForward":"disable","version":"${{ inputs.dotnet-version }}"}}' > global.json
```
duizer, line 149: This not output the correct version. An empty string is outputted.

domains/certificates/RegistryConnector/Worker/Dockerfile
```diff
@@ -1,12 +1,11 @@
...
+FROM mcr.microsoft.com/dotnet/aspnet:${RUNTIME_VERSION}-alpine3.17 AS base
 WORKDIR /app
 EXPOSE 80
 EXPOSE 443
 
-RUN apt-get -y update && apt-get -y install wget
-HEALTHCHECK CMD wget -qO- http://localhost/health > /dev/null || exit 1
-
-FROM mcr.microsoft.com/dotnet/sdk:7.0 AS build
+FROM mcr.microsoft.com/dotnet/sdk:${SDK_VERSION}-alpine3.17 AS build
```
duizer, line 8: This cannot be an alpine image due to our usage of the PO registry client, which does not support MUSL.

.github/workflows/build-libraries.yaml
CodeReaper: It is now.

.github/workflows/build-domain.yaml
CodeReaper: It should be the value from the matrix, yes, and it is now.

domains/certificates/RegistryConnector/Worker/Dockerfile
duizer: Fixed

duizer (approved):

PR #340, opened 2023-06-06
1467 show transfer agreement

ckr123:


domains/transfer-agreements/TransferAgreements.API/API.UnitTests/Controllers/TransferAgreementsControllerTests.cs
```diff
@@ -39,20 +39,29 @@ public TransferAgreementsControllerTests()
 
 
     [Fact]
-    public async Task Create_ShouldCallServiceOnce()
+    public async Task Create_ShouldCallRepositoryOnce()
```
tnickelsen, line 42: I would say these are covered via integration tests? Don't delete these now that we have them :)

tnickelsen (changes requested):

tnickelsen (changes requested):

tnickelsen (dismissed):
Few comments

domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
```diff
@@ -37,4 +39,14 @@ public async Task<ActionResult> Create([FromBody] CreateTransferAgreement reques
 
         return Created($"api/transfer-agreements/{result.Id}", result);
```
duizer, line 40: This has the location of the created resource and the route is hardcoded because, the GET Endpoint was not ready. It is ready now in this PR, so we can get rid of the the hardcoded route. Please consider using CreatedAtAction or CreatedAtRoute here, see this for more info https://ochzhen.com/blog/created-createdataction-createdatroute-methods-explained-aspnet-core

domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
```diff
@@ -37,4 +39,14 @@ public async Task<ActionResult> Create([FromBody] CreateTransferAgreement reques
...
     }
+
+    [HttpGet("{id}")]
+    public async Task<ActionResult> Get([FromRoute] Guid id)
+    {
+        var tin = User.FindFirstValue("tin")!;
+        var subject = User.FindSubjectClaim();
+
+        var result = await transferAgreementRepository.GetTransferAgreement(id, subject, tin);
+        return result == null ? NotFound() : Ok(result);
```
duizer, line 50: The OpenAPI/swagger definition states that this endpoint only returns 200 OK without specifying the type returned.
I suggest you add `ProducesResponseType`-attributes, so it states that it will return 
* 200 OK with specific type returned
* 404 Not Found

duizer, line 50: You are returning the domain entity/business object/database model. Do you not want to map it to a dedicated API response model?

duizer: Also note that the API should communicate all timestamps/dates in Unix Time Seconds

domains/transfer-agreements/TransferAgreements.API/API.IntegrationTests/SwaggerTests.GetSwaggerDoc_AppStarted_NoChangesAccordingToSnapshot.verified.txt
```diff
@@ -65,6 +65,29 @@
           }
         }
       }
+    },
+    "/api/transfer-agreements/{id}": {
```
duizer, line 69: It appears that this snapshot is out-of-sync with the controller

domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
```diff
@@ -42,20 +46,30 @@ public async Task<ActionResult> Create([FromBody] CreateTransferAgreement reques
...
-           StartDate: result.StartDate.ToUnixTimeSeconds(),
-           EndDate: result.EndDate.ToUnixTimeSeconds(),
-           SenderName: result.SenderName,
-           SenderTin: result.SenderTin,
-           ReceiverTin: result.ReceiverTin);
+        return CreatedAtAction(nameof(Get), new { id = result.Id }, ToTransferAgreementDto(result));
+    }
+
+    [ProducesResponseType(typeof(TransferAgreement), 200)]
+    [ProducesResponseType(404)]
```
duizer, line 53: If you change this to 

```suggestion
    [ProducesResponseType(typeof(void), 404)]
```

then the open api definition will say that there is no content returned. The open api definition is currently saying that it returns `ProblemDetails`

duizer (approved):

PR #353, opened 2023-06-23
Preview/1574 data sync syncer new algorithm

tnickelsen:
No end to end tests. There wasn't any before and I am unsure of how they would look like


domains/certificates/Query.API/API/DataSyncSyncer/Persistence/SyncPosition.cs
```diff
@@ -0,0 +1,7 @@
+using System;
+
+namespace API.DataSyncSyncer.Persistence;
+
+
+//Id is required by MartenDb to query: https://martendb.io/documents/identity.html
+public record SyncPosition(Guid Id, string Gsrn, long SyncedTo);
```
ckr123, line 7: Minor: to align with how we usually write Gsrn, it should be GSRN

ckr123 (dismissed):

ckr123 (approved):

PR #370, opened 2023-08-09
CreateWalletDepositEndpoint should propagate existing JWT instead of generating one

sahma19:


domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
```diff
@@ -170,17 +175,16 @@ public async Task<ActionResult<EditTransferAgreementEndDate>> EditEndDate(Guid i
     }
 
     [ProducesResponseType(typeof(string), 200)]
+    [ProducesResponseType(typeof(string), 500)]
```
duizer, line 178: Why is it added that it is expected that the endpoint returns 500 internal server error?

domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
```diff
@@ -170,17 +175,16 @@ public async Task<ActionResult<EditTransferAgreementEndDate>> EditEndDate(Guid i
...
-        var audience = User.FindAudienceClaim();
-        var subject = User.FindSubjectGuidClaim();
-        var name = User.FindSubjectNameClaim();
-
-        var base64String = await walletDepositEndpointService.CreateWalletDepositWithToken(new JwtToken(issuer, audience, subject, name));
-
-        return Ok(new { result = base64String });
+        if (AuthenticationHeaderValue.TryParse(httpContextAccessor.HttpContext?.Request.Headers["Authorization"], out var authentication))
+        {
+            var base64String = await walletDepositEndpointService.CreateWalletDepositWithToken(authentication.Parameter);
```
duizer, line 184: The `WalletDepositEndpointService` explicitly adds (hardodes) "Bearer" infront of the token. Why not just pass the AuthenticationHeaderValue object so that `WalletDepositEndpointService` knows both the scheme and the parameter (the token)?

duizer, line 184: Passing parameters all the way down to the services often leads to hard to read code. Have you considered the option to get the GRPC client from DI? See https://learn.microsoft.com/en-us/aspnet/core/grpc/clientfactory?view=aspnetcore-7.0#call-credentials

Or maybe create a service to be used by `WalletDepositEndpointService`, which is capable of getting the JWT token.

domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
```diff
@@ -170,17 +175,16 @@ public async Task<ActionResult<EditTransferAgreementEndDate>> EditEndDate(Guid i
...
+            return Ok(new { result = base64String });
+        }
+        return StatusCode(500);
```
duizer, line 187: It seems wrong to send a status code 500 due to the fact that there is no jwt-token. Additionally, the controller requires all requests to be authenticated, so there will always be a jwt-token.
I suggest to not use the TryParse-method and instead use the Parse-method: https://learn.microsoft.com/en-us/dotnet/api/system.net.http.headers.authenticationheadervalue.parse?view=net-7.0#system-net-http-headers-authenticationheadervalue-parse(system-string). With this approach there is no branching to "what if we cannot get the jwt token?". Should this fail in the future, it will throw an exception with a saying message and the API client will still get a 500 Status code.

domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
sahma19: A good point. I have actually already made a tech debt issue, on the zenhub board, outlying a need for some refactoring. right now there is a lot of business logic in the controllers, which can be moved to their own respective services, in a service layer. If it is fine with you, I would like to handle it, there?

sahma19: It was to refrain from changing more code than necessary. I will update it, and let you know when it is done

domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
sahma19: Because I am bad at coding :) I'll remove it.

domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
duizer: ok for me

domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
sahma19: Removed

domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
sahma19: Updated so it passes the scheme and token now. Authorization still has to be set in the service since we are using grpc.

domains/transfer-agreements/TransferAgreements.API/API/Controllers/TransferAgreementsController.cs
sahma19: I have switched to parse, assuming the jwt is always correctly formatted. if not it will throw an error.

domains/transfer-agreements/TransferAgreements.API/API/Services/WalletDepositEndpointService.cs
```diff
@@ -56,23 +55,44 @@ public async Task<Guid> CreateReceiverDepositEndpoint(string bearerToken, string
...
+        catch (Exception ex)
         {
-            throw new ArgumentException("The receiver Id cannot be an empty Guid.", nameof(response.ReceiverId));
+            HandleException(ex, "creating ReceiverDepositEndpoint");
+            throw;
         }
+    }
 
-        return receiverReference;
+    private void HandleException(Exception ex, string actionContext)
```
duizer, line 87: Minor thing: In my book, this is not "handling" the exception, only logging it.

domains/transfer-agreements/TransferAgreements.API/API/Services/WalletDepositEndpointService.cs
```diff
@@ -56,23 +55,44 @@ public async Task<Guid> CreateReceiverDepositEndpoint(string bearerToken, string
...
+    {
+        if (ex is RpcException rpcEx)
+        {
+            logger.LogError(rpcEx, "Error from WalletService while {ActionContext}", actionContext);
+        }
+        else
+        {
+            logger.LogError(ex, "Error while {ActionContext}", actionContext);
+        }
```
duizer, line 96: The only difference between these two is the message template. I do not see what value that will bring over just logging the error as on line 95.

Furthermore, what is the point of `actionContext`? There is a stack trace on the exception, so why add this?

duizer (changes requested):

domains/transfer-agreements/TransferAgreements.API/API/Services/WalletDepositEndpointService.cs
sahma19: I removed the method, as your point regarding the difference between exceptions were just the message template, thus rendering it obsolete.

domains/transfer-agreements/TransferAgreements.API/API/Services/WalletDepositEndpointService.cs
sahma19: I removed the method, as your point regarding the difference between exceptions were just the message template, thus rendering it obsolete.

domains/transfer-agreements/TransferAgreements.API/API/API.csproj
```diff
@@ -52,4 +52,8 @@
       <ProjectReference Include="..\..\Shared\ProjectOriginClients\ProjectOriginClients.csproj" />
     </ItemGroup>
 
+    <ItemGroup>
+      <Compile Remove="Data\JwtToken.cs" />
+    </ItemGroup>
```
duizer, line 57: Why is this here?

sahma19: Removed.

duizer (approved):

PR #384, opened 2023-08-17
Certificates: Replace moq with nsubstitute

duizer:


domains/certificates/Query.API/API.IntegrationTests/API.IntegrationTests.csproj
```diff
@@ -16,7 +16,7 @@
     <PackageReference Include="FluentAssertions" Version="6.8.0" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="7.0.4" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.5.0" />
-    <PackageReference Include="Moq" Version="4.18.2" />
```
sahma19, line 19: If the main reason for replacing Moq with NSubsitute, is privacy and security concerns, then I believe having a test, which checks if Moq is present in the project, is a good idea. It would prevent accidental re-introduction, into the project, by another developer as well. The test could be a simple grep that checks for the word "Moq", or a test that checks for package references.

sahma19 (changes requested):

domains/certificates/Query.API/API.IntegrationTests/API.IntegrationTests.csproj
duizer: If we/energy-origin want that feature, it would be a responsibility for the Github action pipeline, which already checks the license for the dependencies. 
But I do not see that as being a part of this PR. You are free to make a change request to Heimdal with that feature.

sahma19 (approved):

PR #390, opened 2023-08-23
One instance of TA

duizer:


domains/transfer-agreements/chart/values.yaml
```diff
@@ -1,5 +1,5 @@
 api:
-  replicaCount: 3
+  replicaCount: 1
```
rvplauborg, line 2: Consider a comment warning about scaling issue :)

ckr123 (approved):

domains/transfer-agreements/chart/values.yaml
duizer: Comment added

ckr123 (approved):

PR #391, opened 2023-08-23
General Container Context view for Energy Origin.

rvplauborg:


doc/diagrams/c4-model/views.dsl
```diff
@@ -35,6 +35,13 @@ workspace extends "https://raw.githubusercontent.com/Energinet-DataHub/opengeh-a
             include ->certificatesDomain->
             autoLayout
         }
+        container energyOrigin "ContainerContextEO" {
+            title "[Container Context] Energy Origin"
+            include *
+            autoLayout
+        }
```
duizer, line 42: What is the reason for having this view? I am not really sure what the point is of this PR...

doc/diagrams/c4-model/views.dsl
```diff
@@ -35,6 +35,13 @@ workspace extends "https://raw.githubusercontent.com/Energinet-DataHub/opengeh-a
...
+
+        # Specific area container views
```
duizer, line 44: Suggest moving the view for Certificates down here

rvplauborg: Will do!

duizer (approved):

PR #398, opened 2023-08-28
Fix sync position for DataSyncSyncer

duizer:
This fixes that SyncPosition is designed in a way, so the underlying table keeps growing for each synchronization. 
When this is released into main, then a follow-up release will be made, which deletes the migration code and tests.


domains/certificates/Query.API/API.IntegrationTests/Repositories/SynchronizationMigrationTests.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        await store.MigrateSynchronizationPosition(CancellationToken.None);
+
+        await using var session = store.QuerySession();
+
+        var syncPositions = await session.Query<SyncPosition>().ToListAsync();
+        syncPositions.Should().HaveCount(0);
+
+        var synchronizationPositions = await session.Query<SynchronizationPosition>().ToListAsync();
+        var expected = gsrns.Select((gsrn, cnt) => new SynchronizationPosition { GSRN = gsrn, SyncedTo = hoursInAYear + cnt });
```
ckr123, line 94: Just by reading this test its hard for me to see whats is expected. Could you be clear in the variable name what it is?

domains/certificates/Query.API/API.IntegrationTests/Repositories/SynchronizationMigrationTests.cs
```diff
@@ -0,0 +1,137 @@
...
+
+    [Fact]
+    public async Task can_be_called_multiple_times()
+    {
+        var gsrn = GsrnHelper.GenerateRandom();
+        var sync1 = new SyncPosition(Guid.NewGuid(), gsrn, 42);
+
+        await SaveSyncPositions(sync1);
+
+        // Call multiple times
```
ckr123, line 52: instead of a comment i would suggest to extract it to a method

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncSyncerWorker.cs
```diff
@@ -150,3 +129,29 @@ private async Task SleepToNearestHour(CancellationToken cancellationToken)
             )
             .ToList();
 }
+
+public static class SynchronizationMigration
+{
+    public static async Task<int> MigrateSynchronizationPosition(this IDocumentStore store, CancellationToken cancellationToken)
```
ckr123, line 135: Is this something temporary just to clean up the db?

domains/certificates/Query.API/API.IntegrationTests/Repositories/SynchronizationMigrationTests.cs
duizer: No. I will remove the comment

domains/certificates/Query.API/API.IntegrationTests/Repositories/SynchronizationMigrationTests.cs
duizer: Sure. Which variable name is it, that should be renamed?

domains/certificates/Query.API/API/DataSyncSyncer/DataSyncSyncerWorker.cs
duizer: Yes. It will be deleted on the next release.

duizer: And so will the corresponding test

ckr123 (approved):

PR #410, opened 2023-09-05
Preview/1731 connections list take2

tnickelsen:


domains/transfer-agreements/TransferAgreements.API/API/Controllers/ConnectionsController.cs
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    private static ConnectionDto ToDto(Connection connection, Guid loggedInCompanyId)
+    {
+        if (loggedInCompanyId == connection.CompanyAId)
+            return new(connection.Id, connection.CompanyBId, connection.CompanyBTin);
+        if (loggedInCompanyId == connection.CompanyBId)
+            return new(connection.Id, connection.CompanyAId, connection.CompanyATin);
+
+        throw new MappingException($"Connection is not owned by the user. Connection: {connection}, logged in companyId: {loggedInCompanyId}");
```
duizer, line 51: I think this situation should never happen, so the question is what is the best approach when it happens.

By doing what the code does now, the request will fail and return 500 Internal Server Error. The pro is that we are most likely going to notice this in our monitoring. The con is that the client is not receiving anything.

Another approach is to log an error and then return null (which is then filtered out in the response to the client). The pro is that the client gets the connections back. The con is that we may ignore the log statement (but that depends on our monitoring setup).

tnickelsen: And I chose Fail Fast and hard (option 1)

PR #418, opened 2023-09-11
fix not working test

ckr123:


domains/transfer-agreements/TransferAgreements.API/API.IntegrationTests/Data/TransferAgreementRepositoryTest.cs
```diff
@@ -3,52 +3,45 @@
...
             SenderTin = "44332211",
             ReceiverTin = "12345678",
             ReceiverReference = Guid.NewGuid(),
-            TransferAgreementNumber = existingTransferAgreement.TransferAgreementNumber
+            TransferAgreementNumber = agreements[0].TransferAgreementNumber
         };
-        dbContext.TransferAgreements.Add(transferAgreement);
-        await Assert.ThrowsAsync<DbUpdateException>(() => dbContext.SaveChangesAsync());
+
+        await Assert.ThrowsAsync<DbUpdateException>(() => factory.SeedTransferAgreementsSaveChangesAsync(transferAgreement));
```
sahma19, line 44: I see that the way you are adding transfer agreements to the db, is directly through seeding the database. Is that the intended approach, or would it be closer to real-life use case, if you tried adding the transferagreements through the API?

domains/transfer-agreements/TransferAgreements.API/API.IntegrationTests/Data/TransferAgreementRepositoryTest.cs
```diff
@@ -57,33 +50,40 @@ private static List<TransferAgreement> SetupTransferAgreements(Guid senderId)
...
+                Id = Guid.NewGuid(),
                 StartDate = DateTimeOffset.UtcNow,
                 EndDate = DateTimeOffset.UtcNow.AddDays(10),
                 SenderId = senderId,
                 SenderName = "nrgi A/S",
                 SenderTin = "44332211",
                 ReceiverTin = "12345678",
                 ReceiverReference = Guid.NewGuid(),
+                TransferAgreementNumber = 3
+
```
sahma19, line 86: The way we are testing here, is without the validations included, which we are skipping because we are seeding the database directly by using SQL. All the validations exist on the service layer. I understand that the goal is to test whether we are triggering the Unique key Contraint, but wouldn't it be better to include validations as well? in a real use case, if a user attempted to create multiple Transfer Agreements, with the same dates, he would be rejected before the request even reaches the repository, let alone the database. Then the Duplicate key violation would never trigger. So this integration test serves to make sure the database is concurrency safe, but data integrity, has not been accounted for, since we are omitting all validations.

domains/transfer-agreements/TransferAgreements.API/API.IntegrationTests/Factories/TransferAgreementsApiWebApplicationFactory.cs
```diff
@@ -81,6 +81,14 @@ public async Task SeedTransferAgreements(IEnumerable<TransferAgreement> transfer
...
     }
 
+    public async Task SeedTransferAgreementsSaveChangesAsync(TransferAgreement transferAgreement)
+    {
+        using var scope = Services.CreateScope();
+        var dbContext = scope.ServiceProvider.GetRequiredService<ApplicationDbContext>();
+        dbContext.TransferAgreements.Add(transferAgreement);
+        await dbContext.SaveChangesAsync();
+    }
+
```
sahma19, line 91: Why is this needed? 

Don't we already have a SeedTransferAgreements method?

sahma19 (approved):

PR #421, opened 2023-09-11
1786 renamed transfer agreement domain to transfer domain

tnickelsen:


domains/transfer/Transfer.API/configuration.yaml
duizer, line 1: The name property in the file says "eo-transfer-agreements-api". It should be renamed. The same goes for the entire domain within eo-base-environement.

duizer (changes requested):
Rename the container image name and also do the name change in eo-base

tnickelsen:
Done in other PR

PR #434, opened 2023-09-14
Deep Links Fix

MortMH:


domains/auth/Auth.API/Unit.Tests/Controllers/OidcControllerTests.cs
```diff
@@ -399,7 +399,8 @@ public async Task CallbackAsync_ShouldReturnRedirectToPath_WhenConfigured()
 
         var uri = new Uri(result.Url);
         Assert.Equal(testOptions.FrontendRedirectUri.Host, uri.Host);
-        Assert.Equal(redirectionPath, uri.AbsolutePath);
+        Assert.Equal(testOptions.FrontendRedirectUri.AbsolutePath, uri.AbsolutePath);
+        Assert.Contains($"redirectionPath={redirectionPath}", HttpUtility.UrlDecode(uri.AbsoluteUri));
```
CodeReaper, line 403: It would be better to extract the actual value from the query string - we do this in some other places when examining 'token' for instance.

CodeReaper (approved):

PR #452, opened 2023-09-27
(Bugfix) GET ConnectionInvitation no longer returns OK, if user is se…

sahma19:
…nder

  - Now returns 400 BadRequest
  - With message "You cannot Accept/Deny your own ConnectionInvitation"


frontend-specialisten (approved):
It looks good to me. The frontend has also tested the changes.

PR #479, opened 2023-10-24
Preview/1824 claim automation

ckr123:


domains/transfer/Transfer.API/API/Shared/Services/ProjectOriginWalletService.cs
```diff
@@ -139,6 +139,25 @@ await walletServiceClient
...
 
+    public async Task<RepeatedField<GranularCertificate>> GetGranularCertificates(Guid subjectId)
+    {
+        var header = SetupDummyAuthorizationHeader(subjectId.ToString());
+        return await GetGranularCertificates(header);
+    }
+
+    public async Task ClaimCertificate(Guid ownerId, GranularCertificate consumptionCertificate, GranularCertificate productionCertificate, uint quantity)
+    {
+        var header = SetupDummyAuthorizationHeader(ownerId.ToString());
```
ckr123, line 150: This could come from some shared context

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
```diff
@@ -0,0 +1,73 @@
...
+
+                        var quantity = productionCertificate.Quantity;
+
+                        if (productionCertificate.Quantity > consumptionCertificate.Quantity)
+                            quantity = consumptionCertificate.Quantity;
+
+                        await walletService.ClaimCertificate(subjectId, consumptionCertificate, productionCertificate, quantity);
+
+                        productionCertificate.Quantity -= quantity;
+                        if (productionCertificate.Quantity <= 0)
```
ckr123, line 54:
```suggestion
if (productionCertificate.Quantity == 0)
```
The check at line `48`should be enough so. we wont allow a larger consumption than production

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
```diff
@@ -0,0 +1,73 @@
...
+            }
+
+            await SleepAnHour(stoppingToken);
+        }
+    }
+
+    private async Task SleepAnHour(CancellationToken cancellationToken)
+    {
+        logger.LogInformation("Sleeping for an hour.");
+        await Task.Delay(TimeSpan.FromMinutes(60), cancellationToken);
```
ckr123, line 71: If it should follow the rest of our background services it should sleep to nearest hours

domains/transfer/Transfer.API/API/Shared/Services/ProjectOriginWalletService.cs
```diff
@@ -16,7 +16,7 @@
 using ProjectOrigin.WalletSystem.V1;
 using Claim = System.Security.Claims.Claim;
 
-namespace API.Transfer.Api.Services;
+namespace API.Shared.Services;
```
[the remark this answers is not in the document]
tnickelsen: Then we DI two services that does the same basically. We can do that but what should we call them? ProjectOriginWalletService1 and ProjectOriginWalletService2?

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
tnickelsen: Yes but is it desired that all background services wake up at the same time so that we have load spikes every hour?

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
tnickelsen: You have faith in my algorithm :D alright!

domains/transfer/Transfer.API/API.UnitTests/Claiming/Automation/ClaimServiceTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        var start = Timestamp.FromDateTimeOffset(DateTimeOffset.Now);
+        var end = Timestamp.FromDateTimeOffset(DateTimeOffset.Now.AddHours(1));
+        uint quantity = 40;
+        var consumptionCertificate1 = BuildCertificate(start, end, GranularCertificateType.Consumption, quantity);
+        var consumptionCertificate2 = BuildCertificate(start, end, GranularCertificateType.Consumption, quantity);
+        var productionCertificate = BuildCertificate(start, end, GranularCertificateType.Production, quantity);
+        using var cts = new CancellationTokenSource();
+
+        claimRepository.GetClaimSubjects().Returns(new List<ClaimSubject> { claimSubject });
+        poWalletService.GetGranularCertificates(Arg.Any<Guid>()).Returns(new RepeatedField<GranularCertificate>
```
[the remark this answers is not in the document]
tnickelsen: How would that look like? Then I would have a function with a big touple return type or something like it - I don't like that

domains/transfer/Transfer.API/API/Shared/Services/ProjectOriginWalletService.cs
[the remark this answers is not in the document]
ckr123: No need for different names, it is in different namespaces with different startup files. Alternative this should be a seperate service

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
ckr123: dno. Its fine by me to keep it as it is.

domains/transfer/Transfer.API/API/Claiming/Automation/Services/IProjectOriginWalletService.cs
```diff
@@ -0,0 +1,12 @@
+using Google.Protobuf.Collections;
+using ProjectOrigin.WalletSystem.V1;
+using System.Threading.Tasks;
+using System;
+
+namespace API.Claiming.Automation.Services;
+
+public interface IProjectOriginWalletService
+{
+    Task<RepeatedField<GranularCertificate>> GetGranularCertificates(Guid subjectId);
```
duizer, line 10: Why use RepeatedField here not use IEnumerable?

domains/transfer/Transfer.API/API/Shared/Services/ProjectOriginService.cs
```diff
@@ -0,0 +1,25 @@
+using Grpc.Core;
+using Microsoft.IdentityModel.Tokens;
+using System.IdentityModel.Tokens.Jwt;
+using System.Security.Claims;
+using System;
+
+namespace API.Shared.Services;
+
+public abstract class ProjectOriginService
```
duizer, line 9: Not really sure if this should be a base class at all and not just a static helper method. But if kept as a base class, the name "Wallet" should be added to indicate that this is for the Wallet.

domains/transfer/Transfer.API/API.IntegrationTests/Claiming/Api/Controllers/ClaimControllerTest.cs
```diff
@@ -0,0 +1,154 @@
...
+        });
+
+        var client = factory.CreateAuthenticatedClient(sub: subject.ToString());
+        var result = await client.GetFromJsonAsync<ClaimSubjectHistoryEntriesDto>("api/claims/claim-process-history");
+
+        result.Should().NotBeNull();
+        result!.History.Should().HaveCount(1);
+        result.History[0].ActorName.Should().Be(historyEntry.ActorName);
+        result.History[0].CreatedAt.Should().BeSameDateAs(historyEntry.CreatedAt);
+        result.History[0].AuditAction.Should().Be(historyEntry.AuditAction);
```
duizer, line 124: Have you considered testing this by not directly inserting data into the history-table? This does not really test if the history entries are generated when doing changes on the API.

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimController.cs
```diff
@@ -0,0 +1,108 @@
...
+using API.Shared.Extensions;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.EntityFrameworkCore;
+
+namespace API.Claiming.Api.Controllers;
+
+[Authorize]
+[ApiController]
+[Route("api/claims")]
```
duizer, line 17: We have decided to not a "root" entity in our API paths. My suggestion is to remove this and then have e.g. 

`GET /api/claim-process`
`POST /api/claim-process/start`
`DELETE /api/claim-process/stop` or `POST /api/claim-process/stop` or `DELETE /api/claim-process`

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
```diff
@@ -0,0 +1,79 @@
...
+            var quantity = productionCertificate.Quantity;
+
+            if (productionCertificate.Quantity > consumptionCertificate.Quantity)
+                quantity = consumptionCertificate.Quantity;
+
+            await walletService.ClaimCertificate(subjectId, consumptionCertificate, productionCertificate, quantity);
+
+            productionCertificate.Quantity -= quantity;
+            if (productionCertificate.Quantity == 0)
+                productionCertificates.Remove(productionCertificate);
```
duizer, line 70: Have you considered that claiming is an asynchronous operation in the Wallet+Registry? And if it is a good idea to send multiple claim operations against the same production certificate but different consumption certificates?

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            var productionCertificate = productionCertificates.FirstOrDefault(x =>
+                x.Start == consumptionCertificate.Start && x.End == consumptionCertificate.End &&
+                x.GridArea == consumptionCertificate.GridArea);
+            if (productionCertificate == null) continue;
+
+            var quantity = productionCertificate.Quantity;
+
+            if (productionCertificate.Quantity > consumptionCertificate.Quantity)
+                quantity = consumptionCertificate.Quantity;
```
duizer, line 64: Could just use Math.Min() for this

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+    }
+
+    private async Task Claim(Guid subjectId, List<GranularCertificate> consumptionCertificates, List<GranularCertificate> productionCertificates)
+    {
+        foreach (var consumptionCertificate in consumptionCertificates)
+        {
+            var productionCertificate = productionCertificates.FirstOrDefault(x =>
+                x.Start == consumptionCertificate.Start && x.End == consumptionCertificate.End &&
+                x.GridArea == consumptionCertificate.GridArea);
```
duizer, line 58: With this logic a consumption cert will only be used in a single claim, but a production cert can be used multiple times. Why?

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
```diff
@@ -0,0 +1,79 @@
...
+
+    public async Task Run(CancellationToken stoppingToken)
+    {
+        while (!stoppingToken.IsCancellationRequested)
+        {
+            logger.LogInformation("ClaimService running at: {time}", DateTimeOffset.Now);
+            try
+            {
+                var claimSubjects = await claimRepository.GetClaimSubjects();
+                foreach (var subjectId in claimSubjects.Select(x => x.SubjectId))
```
duizer, line 34: Reading this I thought: "Does this run multiple times for the same SubjectId, if it is repeated in `claimSubject`"?

You could use .Distinct() in stead of .Select(). Since SubjectId is the key, it will give the same result, but reading this will be easier to see that this is only unique SubjectIds.

domains/transfer/Transfer.API/API.IntegrationTests/Claiming/Api/Controllers/ClaimControllerTest.cs
ckr123: When doing so i am running into some flaky test where history isnt always  created before making the request - do you have any ideas on how to solve it?

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimController.cs
ckr123: out of curiosity didnt we decide to remove the topic from the guldelines and let the teams decide?

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
tnickelsen: No. At least not more than what I'm doing - only claiming quantity that is 'left' of the certificate

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
tnickelsen: I don't follow

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
tnickelsen: I can't get it to look as elegant using destinct. Can you show me what you mean?

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
tnickelsen: Good point. For no other reason than that I needed to enumerate one of them. How would you suggest doing it so that the consumption certificate can be used multiple times as well?

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
ckr123: Always :)

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
duizer: var quantity = Math.Min(productionCertificate.Quantity, consumptionCertificate.Quantity);

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimController.cs
```diff
@@ -0,0 +1,108 @@
...
+public class ClaimController : ControllerBase
+{
+    private readonly IClaimRepository claimRepository;
+
+    public ClaimController(IClaimRepository claimRepository)
+    {
+        this.claimRepository = claimRepository;
+    }
+
+    [HttpPost("start-claim-process")]
```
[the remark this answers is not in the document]
duizer: Let's go with `claim-automation` as the name for the resource. This specific endpoint should be `POST /api/claim-automation/start`

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
duizer: claimSubjects.Select(x => x.SubjectId).Distinct()

domains/transfer/Transfer.API/API.IntegrationTests/Claiming/Api/Controllers/ClaimControllerTest.cs
ckr123: fixed eb582eb2

ckr123: and e48a3de2

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimController.cs
ckr123: fixed eb582eb

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimController.cs
```diff
@@ -0,0 +1,108 @@
...
+        if (claim == null)
+        {
+            return NotFound();
+        }
+
+        claimRepository.DeleteClaimSubject(claim);
+        return NoContent();
+    }
+
+    [HttpGet("claim-process-history")]
```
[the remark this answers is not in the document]
ckr123: fixed eb582eb

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimController.cs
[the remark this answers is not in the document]
ckr123: Fixed eb582eb

domains/transfer/Transfer.API/API/Claiming/Api/Dto/Response/ClaimSubjectHistoryEntriesDto.cs
```diff
@@ -0,0 +1,6 @@
+using System.Collections.Generic;
+using API.Claiming.Api.Models;
+
+namespace API.Claiming.Api.Dto.Response;
+
+public record ClaimSubjectHistoryEntriesDto(List<ClaimSubjectHistoryEntryDto> History);
```
duizer, line 6: Does this follow the rules of pagination in the API Guidelines? Also, it differs from the same history response for Transfer Agreement

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimController.cs
```diff
@@ -0,0 +1,107 @@
...
+    public async Task<ActionResult<ClaimSubject>> GetClaimProcess()
+    {
+        var subject = User.FindSubjectGuidClaim();
+        var claim = await claimRepository.GetClaimSubject(Guid.Parse(subject));
+        if (claim == null)
+        {
+            return NotFound();
+        }
+
+        return Ok(claim);
```
duizer, line 105: Have you considered returning when the claim automation was started?

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimController.cs
```diff
@@ -0,0 +1,107 @@
...
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.EntityFrameworkCore;
+
+namespace API.Claiming.Api.Controllers;
+
+[Authorize]
+[ApiController]
+[Route("api/claim-automation")]
+public class ClaimController : ControllerBase
```
duizer, line 17: I suggest to call this controller for `ClaimAutomationController`. Also, the methods name uses the term "claim process" and not "claim automation"

domains/transfer/Transfer.API/API.IntegrationTests/Claiming/Api/Controllers/ClaimControllerTest.cs
```diff
@@ -0,0 +1,132 @@
...
+        var client = factory.CreateAuthenticatedClient(sub: Guid.NewGuid().ToString(), actor: actor);
+        await client.PostAsync("api/claim-automation/start", null);
+        var result = await client
+            .RepeatedlyGetUntil<ClaimSubjectHistoryEntriesDto>(
+                "api/claim-automation/history",
+                res => res.History.Any()
+            );
+
+        result.Should().NotBeNull();
+        result.History[0].ActorName.Should().Be(actor);
```
duizer, line 102: It would make sense to assert on the entire resonse and not just a single property on the first history entry. You can use verfiy.xunit to make this a snapshot test

domains/transfer/Transfer.API/API.IntegrationTests/Claiming/Api/Controllers/ClaimControllerTest.cs
```diff
@@ -0,0 +1,132 @@
...
+        await client1.PostAsync("api/claim-automation/start", null);
+
+        var client = factory.CreateAuthenticatedClient(sub: Guid.NewGuid().ToString());
+        var result = await client.GetAsync("api/claim-automation/history");
+
+        result.StatusCode.Should().Be(HttpStatusCode.NotFound);
+    }
+
+    [Fact]
+    public async Task GetClaimSubjectHistory_WhenHistoryExists_ReturnsOK()
```
duizer, line 90: I would like that there is a history test for starting claim automation, stopping and then starting again. In that case, there should be 3 entries in the history response.

domains/transfer/Transfer.API/API/Shared/Services/ProjectOriginWalletService.cs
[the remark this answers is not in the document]
duizer: The only thing that is shared is a static helper method. You do not need to have a base class. A static helper class with a static method is all that needs to be shared.

domains/transfer/Transfer.API/API/Claiming/Api/Dto/Response/ClaimSubjectHistoryEntriesDto.cs
ckr123: It doesnt follow pagination. 
Maybe it should.
Regarding the response type its purpose is to serve the frontend and they havent specifiec the needs yet.do you refer to the enum on the auditaction?

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimController.cs
ckr123: No i havent. The frontend is not using the timestamp

domains/transfer/Transfer.API/API/Claiming/Api/Dto/Response/ClaimSubjectHistoryEntriesDto.cs
duizer: I was referred to the name of the property for the array. In the the API guidelines, it is called `items`: https://energinet.atlassian.net/wiki/spaces/EnergyOrigin/pages/460947457/API+Guidelines#Must-support-pagination. For this endpoint, it is not called `items`, but is called `history`. For TA history, it is called `result`.

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimController.cs
duizer: This is returning an object of type `ClaimSubject`, which only has one property and that is the subject id. Technically, the frontend does not even need the subject id returned. 
But I just thought it would make sense to at least store a timestamp for when the claim automation was started.

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
duizer: Sorry, you can also use .DistinctBy(). See https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/linq/set-operations?branch=main#distinct-and-distinctby

tnickelsen: Yes but then I get ClaimSubjects and have to Select afterswards anyway

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
```diff
@@ -0,0 +1,81 @@
...
+
+            await SleepAnHour(stoppingToken);
+        }
+    }
+
+    private async Task Claim(Guid subjectId, List<GranularCertificate> consumptionCertificates, List<GranularCertificate> productionCertificates)
+    {
+        while (productionCertificates.Any(x => x.Quantity > 0) && consumptionCertificates.Any(x => x.Quantity > 0))
+        {
+            var productionCert = productionCertificates.FirstOrDefault(x => x.Quantity > 0);
```
duizer, line 62: Null checks are not needed here due to the condition in the while-loop

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
```diff
@@ -0,0 +1,81 @@
...
+                foreach (var subjectId in claimSubjects.Select(x => x.SubjectId).Distinct())
+                {
+                    var certificates = await walletService.GetGranularCertificates(subjectId);
+
+                    var certificatesGrouped = certificates.GroupBy(x => new { x.GridArea, x.Start, x.End });
+
+                    foreach (var cert in certificatesGrouped)
+                    {
+                        var productionCerts = cert.Where(x => x.Type == GranularCertificateType.Production).ToList();
+                        var consumptionCerts = cert.Where(x => x.Type == GranularCertificateType.Consumption).ToList();
```
duizer, line 43: The order of the certificates may be fixed by the wallet (e.g. they are ordered by issuance date), which with this claiming/matching algorithm will introduce a bias to which certificates are matched/claimed first.
I will suggest that the certificates are shuffled/ordered randomly at some point to ensure that this algorithm is not biased.

domains/transfer/Transfer.API/API.UnitTests/Claiming/Automation/ClaimServiceTests.cs
```diff
@@ -0,0 +1,131 @@
...
+
+namespace API.UnitTests.Claiming.Automation;
+
+public class ClaimServiceTests
+{
+    private static class Attributes
+    {
+        public const string AssetId = "AssetId";
+        public const string TechCode = "TechCode";
+        public const string FuelCode = "FuelCode";
```
duizer, line 26: This is only used once in this test, so it can be inlined.

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimAutomationController.cs
```diff
@@ -0,0 +1,102 @@
...
+        var history = await claimRepository.GetHistory(Guid.Parse(subject));
+
+        if (history.Count == 0)
+        {
+            return NotFound();
+        }
+
+        var historyDto = history.Select(c =>
+                c.ToDto()
+            ).ToList();
```
duizer, line 83: Indentation may be valid here, but it looks strange.

domains/transfer/Transfer.API/API/Claiming/Api/Dto/Response/ClaimSubjectHistoryEntryDto.cs
```diff
@@ -0,0 +1,35 @@
...
+
+public record ClaimSubjectHistoryEntryDto(
+    long CreatedAt,
+    ChangeAction Action,
+    string? ActorName
+);
+
+public static class ClaimSubjectHistoryEntryMapper
+{
+    public static ClaimSubjectHistoryEntryDto ToDto(this ClaimSubjectHistory historyEntry)
```
duizer, line 15: Here you map a "model" to a "DTO" (I am trying to use the wording based on the folder-names containing the classes). But that is not the case for `ClaimSubject`, where the "model" is returned directly by the API and there is not "DTO". Why?

domains/transfer/Transfer.API/API/Claiming/Api/Models/ClaimSubject.cs
```diff
@@ -0,0 +1,5 @@
+using System;
+
+namespace API.Claiming.Api.Models;
+
+public record ClaimSubject(Guid SubjectId);
```
duizer, line 5: I would prefer that the timestamp for when this entity was created was saved. It may be needed later and is not possible to recreate at a later point.

domains/transfer/Transfer.API/API/Claiming/Api/Repositories/ClaimRepository.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using API.Claiming.Api.Models;
+using API.Shared.Data;
+using Microsoft.EntityFrameworkCore;
+
+namespace API.Claiming.Api.Repositories;
+
+public class ClaimRepository : IClaimRepository
```
duizer, line 11: Depending on how strong you feel about this, I would rename this to `ClaimAutomationRepository` as it is not a repository for claims.

domains/transfer/Transfer.API/API/Claiming/Api/Repositories/IClaimRepository.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using API.Claiming.Api.Models;
+
+namespace API.Claiming.Api.Repositories;
+
+public interface IClaimRepository
+{
+    Task<List<ClaimSubject>> GetClaimSubjects();
+    Task<ClaimSubject?> GetClaimSubject(Guid subject);
```
duizer, line 11: Not sure I have a much better name, but I think the name `ClaimSubject` is confusing. Or it does not express the intent.

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
```diff
@@ -0,0 +1,81 @@
...
+
+            productionCert.Quantity -= quantity;
+            consumptionCert.Quantity -= quantity;
+        }
+    }
+
+    private async Task SleepAnHour(CancellationToken cancellationToken)
+    {
+        logger.LogInformation("Sleeping for an hour.");
+        await Task.Delay(TimeSpan.FromMinutes(60), cancellationToken);
```
duizer, line 79: By doing it this way, the service will not run every 1 hour, but will run every 1 hour + the time it takes to do the claiming.

domains/transfer/Transfer.API/API/Claiming/Automation/Services/ProjectOriginWalletService.cs
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public async Task<List<GranularCertificate>> GetGranularCertificates(Guid subjectId)
+    {
+        var header = ProjectOriginWalletHelper.SetupDummyAuthorizationHeader(subjectId.ToString());
+        var response = await walletServiceClient.QueryGranularCertificatesAsync(new QueryRequest(), header);
+        return response.GranularCertificates.ToList();
+    }
+
+    public async Task ClaimCertificate(Guid ownerId, GranularCertificate consumptionCertificate, GranularCertificate productionCertificate, uint quantity)
```
duizer, line 26:
```suggestion
    public async Task ClaimCertificates(Guid ownerId, GranularCertificate consumptionCertificate, GranularCertificate productionCertificate, uint quantity)
```

domains/transfer/Transfer.API/API/Shared/Data/ApplicationDbContext.cs
```diff
@@ -25,5 +28,13 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
         modelBuilder.Entity<TransferAgreement>()
             .HasIndex(nameof(TransferAgreement.SenderId), nameof(TransferAgreement.TransferAgreementNumber))
             .IsUnique();
+
+        modelBuilder.Entity<ClaimSubject>()
+            .HasKey(p => p.SubjectId);
+
+        modelBuilder.Entity<ClaimSubjectHistory>()
+            .HasKey(p => p.SubjectId);
```
duizer, line 36: This defines the primary key to be "SubjectId", but that will mean, that it is not possible to have multiple entries for the same subject id. Is this a mistake?

domains/transfer/Transfer.API/API/Program.cs
```diff
@@ -85,9 +92,66 @@
...
+                        break;
+                    case "Update":
+                        {
+                            historyEntity.TransferAgreementId = (Guid)eventEntry.PrimaryKey.Values.First();
+                            break;
+                        }
+                }
+
+                return true;
+            })
```
duizer, line 135: Could this be moved into the startup-file for TransferAgreement? The same goes for the claim part below

domains/transfer/Transfer.API/API/Transfer/Startup.cs
```diff
@@ -22,53 +23,26 @@ public static class Startup
...
                 provider
-                    .SetResourceBuilder(ResourceBuilder.CreateDefault().AddService(TransferAgreementAutomationMetrics.MetricName))
+                    .SetResourceBuilder(ResourceBuilder.CreateDefault()
+                        .AddService(TransferAgreementAutomationMetrics.MetricName))
                     .AddMeter(TransferAgreementAutomationMetrics.MetricName)
                     .AddHttpClientInstrumentation()
                     .AddAspNetCoreInstrumentation()
                     .AddRuntimeInstrumentation()
                     .AddProcessInstrumentation()
                     .AddOtlpExporter(o => o.Endpoint = otlpOptions.ReceiverEndpoint));
```
duizer, line 45: This is not specific for TransferAgreement automation, but applies for the entire application.

domains/transfer/Transfer.API/API/Program.cs
```diff
@@ -85,9 +92,66 @@
...
 
+
 builder.Services.AddTransfer(builder.Configuration);
 builder.Services.AddCvr();
 builder.Services.AddConnection();
+builder.Services.AddClaimServices();
+
+builder.Services.AddControllers(options => options.Filters.Add<AuditDotNetFilter>())
+    .AddJsonOptions(options =>
+        options.JsonSerializerOptions.Converters.Add(new JsonStringEnumConverter()));
```
duizer, line 103: There is a class named `JsonDefault` which adds the same converter (with a small difference). Why is `JsonDefault` not used here?

domains/transfer/Transfer.API/API.IntegrationTests/Claiming/Api/Controllers/ClaimAutomationControllerTest.cs
```diff
@@ -0,0 +1,157 @@
...
+        await Verifier.Verify(result, settings);
+    }
+
+    [Fact]
+    public async Task GetClaimSubjectHistory_WhenMultipleHistoryExists_ReturnsOK()
+    {
+        const string actor = "Peter Producent";
+        var client = factory.CreateAuthenticatedClient(sub: Guid.NewGuid().ToString(), actor: actor);
+        await client.PostAsync("api/claim-automation/start", null);
+        await Task.Delay(5000);
```
duizer, line 115: Why is this delay needed? If it is a work-around, could a different solution be to change the condition in "RepeatedlyGetUntil" to exactly 2 entries instead of "Any"?

domains/transfer/Transfer.API/API.IntegrationTests/Claiming/Api/Controllers/ClaimControllerTest.GetClaimSubjectHistory_WhenMultipleHistoryExists_ReturnsOK.verified.txt
```diff
@@ -0,0 +1,14 @@
+﻿{
+  History: [
+    {
+      CreatedAt: {Scrubbed},
+      Action: Created,
+      ActorName: Peter Producent
+    },
+    {
+      CreatedAt: {Scrubbed},
+      Action: Created,
```
duizer, line 10: I would expect this to be a deletion.

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimAutomationController.cs
```diff
@@ -0,0 +1,102 @@
...
+    {
+        var subject = User.FindSubjectGuidClaim();
+        var history = await claimRepository.GetHistory(Guid.Parse(subject));
+
+        if (history.Count == 0)
+        {
+            return NotFound();
+        }
+
+        var historyDto = history.Select(c =>
```
duizer, line 81: It will be a good idea to sort the history entries by the timestamp.

domains/transfer/Transfer.API/API.IntegrationTests/API.IntegrationTests.csproj
```diff
@@ -52,5 +52,11 @@
 		<None Update="Cvr\Api\Controllers\CvrControllerTests.empty_cvr_response.json">
 		  <CopyToOutputDirectory>Always</CopyToOutputDirectory>
 		</None>
+		<None Update="Claiming\Api\Controllers\ClaimControllerTest.GetClaimSubjectHistory_WhenHistoryExists_ReturnsOK.verified.txt">
+		  <DependentUpon>ClaimAutomationControllerTest.cs</DependentUpon>
+		</None>
+		<None Update="Claiming\Api\Controllers\ClaimControllerTest.GetClaimSubjectHistory_WhenMultipleHistoryExists_ReturnsOK.verified.txt">
+		  <DependentUpon>ClaimAutomationControllerTest.cs</DependentUpon>
+		</None>
```
duizer, line 60: This is normally not needed when working with verify.xunit.

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimWorker.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        this.serviceProvider = serviceProvider;
+    }
+
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        using var scope = serviceProvider.CreateScope();
+        var claimService = scope.ServiceProvider.GetRequiredService<IClaimService>();
+
+        await claimService.Run(stoppingToken);
```
duizer, line 23: When resolving this way, there is only created one instance of IClaimService that lives the entire application lifetime. Following the line of dependencies for ClaimService, we hit a DbContext; this instance of DbContext is never disposed and is against the recommendation. See https://learn.microsoft.com/en-us/ef/core/dbcontext-configuration/#the-dbcontext-lifetime.
You should use IDbContextFactory instead to create an instance of the DbContext when needed (which is only on the start of each run for claim automation).

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimAutomationController.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        try
+        {
+            claim = await claimRepository.AddClaimSubject(claimSubject);
+            return CreatedAtAction(nameof(GetClaimAutomation), null, claim);
+        }
+        catch (DbUpdateException)
+        {
+            return CreatedAtAction(nameof(GetClaimAutomation), null, claimSubject);
+        }
```
duizer, line 49: Why is this needed?

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
tnickelsen: corrected

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
tnickelsen: Why do you suggest that we actively do something to shuffle it? What do we gain?

domains/transfer/Transfer.API/API.UnitTests/Claiming/Automation/ClaimServiceTests.cs
tnickelsen: corrected

domains/transfer/Transfer.API/API/Claiming/Api/Models/ClaimSubject.cs
tnickelsen: Added CreatedAt and migrations

domains/transfer/Transfer.API/API/Claiming/Api/Repositories/ClaimRepository.cs
tnickelsen: corrected

domains/transfer/Transfer.API/API/Claiming/Api/Repositories/ClaimRepository.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    private readonly ApplicationDbContext context;
+
+    public ClaimRepository(ApplicationDbContext context)
+    {
+        this.context = context;
+    }
+
```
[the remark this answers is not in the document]
tnickelsen: corrected

domains/transfer/Transfer.API/API/Claiming/Api/Repositories/IClaimRepository.cs
tnickelsen: It was lack of a better name

domains/transfer/Transfer.API/API/Claiming/Automation/Services/ProjectOriginWalletService.cs
tnickelsen: corrected

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
tnickelsen: Good point. We could set it to every hour staring at half (ie 1.30, 2.30, 3.30, etc). Was just trying to avoid everything waking up at the same time

domains/transfer/Transfer.API/API/Shared/Data/ApplicationDbContext.cs
tnickelsen: No since you can only start the robot once pr subjectId

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimWorker.cs
tnickelsen: I have corrected this. Let me know if I did it wrong

domains/transfer/Transfer.API/API/Program.cs
ckr123: You cannot have two configurations of audit.net. Dno how to split them, got any ideas?

domains/transfer/Transfer.API/API/Claiming/Api/Dto/Response/ClaimSubjectHistoryEntriesDto.cs
tnickelsen: corrected to items

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimController.cs
tnickelsen: Added CreatedAt to ClaimSubject so it now returns timestamp as well

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimController.cs
tnickelsen: This has been corrected

domains/transfer/Transfer.API/API.IntegrationTests/Claiming/Api/Controllers/ClaimControllerTest.cs
tnickelsen: Now uses verify

domains/transfer/Transfer.API/API.IntegrationTests/Claiming/Api/Controllers/ClaimControllerTest.cs
tnickelsen: Added

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimAutomationController.cs
tnickelsen: corrected

domains/transfer/Transfer.API/API/Shared/Data/ApplicationDbContext.cs
duizer: It does not make sense to set the primary key for ClaimSubjectHistory to SubjectId, when there is also an Id property.

domains/transfer/Transfer.API/API/Program.cs
duizer: You can use an extension method for that

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
duizer: We would ensure that it is actually shuffled - this means that there is no bias in the algorithm.

duizer: Why do you suggest that we do not shuffle it? Why do you want to base the matching on the order of the certificates returned by the Wallet?

domains/transfer/Transfer.API/API/Claiming/Api/Dto/Response/ClaimSubjectHistoryEntryDto.cs
tnickelsen: Added dto for ClaimSubject

domains/transfer/Transfer.API/API/Transfer/Startup.cs
tnickelsen: Moved to program

domains/transfer/Transfer.API/API/Program.cs
tnickelsen: To be removed

domains/transfer/Transfer.API/API.IntegrationTests/Claiming/Api/Controllers/ClaimAutomationControllerTest.cs
tnickelsen: To be removed

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimAutomationController.cs
tnickelsen: To be removed

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimAutomationController.cs
tnickelsen: To catch if a subject with same primary key already exists

domains/transfer/Transfer.API/API/Shared/Services/ProjectOriginWalletService.cs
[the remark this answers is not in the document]
tnickelsen: Had been added

domains/transfer/Transfer.API/API/Claiming/Automation/ClaimService.cs
tnickelsen: Because we actively have to shuffle it - this looks cleaner. Isn't it ok with bias? Then we know what certificates it takes first

domains/transfer/Transfer.API/API/Shared/Data/ApplicationDbContext.cs
tnickelsen: Sorry - thought we were talking about ClaimSubject. Agreed, this does not make sense. But this is to be removed so doesn't matter anymore

domains/transfer/Transfer.API/API/Program.cs
ckr123: Are you referring to the jsondefault from integration test?

domains/transfer/Transfer.API/API/Shared/Migrations/20231101131822_UpdateClaimSubjectWithCreatedAt.cs
```diff
@@ -0,0 +1,30 @@
+using System;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+#nullable disable
+
+namespace API.Shared.Migrations
+{
+    /// <inheritdoc />
+    public partial class UpdateClaimSubjectWithCreatedAt : Migration
```
duizer, line 9: Can we have just one migration for this single table?

domains/transfer/Transfer.API/API.IntegrationTests/SwaggerTests.GetSwaggerDoc_AppStarted_NoChangesAccordingToSnapshot.verified.txt
```diff
@@ -5,6 +5,136 @@
...
+        "tags": [
+          "ClaimAutomation"
+        ],
+        "responses": {
+          "201": {
+            "description": "Created",
+            "content": {
+              "text/plain": {
+                "schema": {
+                  "$ref": "#/components/schemas/ClaimSubject"
```
duizer, line 19: ClaimSubjectDto

domains/transfer/Transfer.API/API.IntegrationTests/SwaggerTests.GetSwaggerDoc_AppStarted_NoChangesAccordingToSnapshot.verified.txt
```diff
@@ -755,6 +885,34 @@
...
+        },
+        "additionalProperties": false
+      },
+      "ClaimSubjectDto": {
+        "type": "object",
+        "properties": {
+          "subjectId": {
+            "type": "string",
+            "format": "uuid"
+          },
```
duizer, line 908: Do not return this in the API

domains/transfer/Transfer.API/API.IntegrationTests/Claiming/Api/Controllers/ClaimAutomationControllerTest.cs
```diff
@@ -0,0 +1,100 @@
...
+        });
+
+        var client = factory.CreateAuthenticatedClient(sub: subject.ToString());
+
+        var result = await client.PostAsync("api/claim-automation/start", null);
+        result.StatusCode.Should().Be(HttpStatusCode.OK);
+    }
+
+    [Fact]
+    public async Task GetClaimSubjects_WhenClaimSubjectsExists_ReturnsOK()
```
duizer, line 74: The term ClaimSubject is obsolete and should be renamed.

domains/transfer/Transfer.API/API/Program.cs
duizer: It is not the same. Closing

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimAutomationController.cs
```diff
@@ -0,0 +1,87 @@
...
+        var claimSubject = new ClaimAutomationArgument(Guid.Parse(subject), DateTimeOffset.UtcNow);
+
+        try
+        {
+            claim = await claimAutomationRepository.AddClaimSubject(claimSubject);
+            var claimSubjectDto = new ClaimAutomationArgumentDto(claim.CreatedAt);
+
+            return CreatedAtAction(nameof(GetClaimAutomation), null, claimSubjectDto);
+        }
+        catch (DbUpdateException)
```
duizer, line 48: This may catch all types of Database errors and not only if the record already exists (which will be a primary key violation).

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimAutomationController.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            var claimSubjectDto = new ClaimAutomationArgumentDto(claimSubject.CreatedAt);
```
duizer, line 50: Should refetch from the database here to return exact information for claim Automation Argument

domains/transfer/Transfer.API/API.IntegrationTests/SwaggerTests.GetSwaggerDoc_AppStarted_NoChangesAccordingToSnapshot.verified.txt
```diff
@@ -5,6 +5,119 @@
...
+        "responses": {
+          "204": {
+            "description": "No Content"
+          },
+          "404": {
+            "description": "Not Found",
+            "content": {
+              "text/plain": {
+                "schema": {
+                  "$ref": "#/components/schemas/ProblemDetails"
```
duizer, line 71: Should not be a ProblemDetails here.

domains/transfer/Transfer.API/API/Claiming/Api/Dto/Response/ClaimAutomationArgumentDto.cs
```diff
@@ -0,0 +1,5 @@
+using System;
+
+namespace API.Claiming.Api.Dto.Response;
+
+public record ClaimAutomationArgumentDto(DateTimeOffset CreatedAt);
```
duizer, line 5: This should be in UnixTimeSeconds

ckr123: long?

ckr123: fixed 7d3053de

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimAutomationController.cs
ckr123: fixed

domains/transfer/Transfer.API/API/Claiming/Api/Controllers/ClaimAutomationController.cs
ckr123: fixed

domains/transfer/Transfer.API/API.IntegrationTests/Claiming/Api/Controllers/ClaimAutomationControllerTest.cs
ckr123: fixed

domains/transfer/Transfer.API/API/Claiming/Startup.cs
```diff
@@ -0,0 +1,21 @@
...
+
+public static class Startup
+{
+    public static void AddClaimServices(this IServiceCollection services)
+    {
+        services.AddScoped<IClaimAutomationRepository, ClaimAutomationRepository>();
+
+        services.AddScoped<IClaimService, ClaimService>();
+        services.AddScoped<IProjectOriginWalletService, ProjectOriginWalletService>();
+        services.AddScoped<Random>();
```
duizer, line 17: I find it quite surprising to see the .NET class Random being registered in the DI. If you want to do it that way, I would prefer that you create a new class - let's call it "Shuffler" - that wraps an instance of Random. The "Shuffler" class is registered in DI, not Random.

domains/transfer/Transfer.API/API/Claiming/Api/Repositories/ClaimAutomationRepository.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        await using var context = await contextFactory.CreateDbContextAsync();
+
+        await context.ClaimAutomationArguments.AddAsync(claimAutomationArgument);
+        await context.SaveChangesAsync();
+
+        return claimAutomationArgument;
+    }
+
+    public async void DeleteClaimAutomationArgument(ClaimAutomationArgument claim)
```
duizer, line 44: Please do not do `async void`
```suggestion
    public async Task DeleteClaimAutomationArgument(ClaimAutomationArgument claim)
```

domains/transfer/Transfer.API/API.IntegrationTests/Claiming/Api/Controllers/ClaimAutomationControllerTest.cs
```diff
@@ -0,0 +1,100 @@
...
+        var claimAutomationArgument = new ClaimAutomationArgument(subject, DateTimeOffset.UtcNow);
+
+        await factory.SeedClaims(new List<ClaimAutomationArgument>()
+        {
+            claimAutomationArgument
+        });
+
+        var client = factory.CreateAuthenticatedClient(sub: subject.ToString());
+        var result = await client.GetAsync("api/claim-automation/");
+        result.EnsureSuccessStatusCode();
```
duizer, line 87: The other tests assert on the exact status code: 
```suggestion
        result.StatusCode.Should().Be(HttpStatusCode.OK);
```

PR #487, opened 2023-11-03
Certificates issuer shared data access

duizer:
Preview env: https://eo-u-certissuershar.westeurope.cloudapp.azure.com/


domains/certificates/README.md
```diff
@@ -64,11 +64,6 @@ You must manually remember to generate the complete SQL migration script after a
 This is the command for generating the migration SQL script for the API project:
 
 ```shell
-dotnet ef migrations script --idempotent --project Query.API/API/API.csproj --output migrations/API.sql
+dotnet ef migrations script --idempotent --project Query.API/API --output migrations/API.sql
+cp migrations/API.sql migrations/Worker.sql
```
tnickelsen, line 68: Why not set the output to migrations/Worker.sql?

domains/certificates/Shared/DataContext/Migrations/ApplicationDbContextModelSnapshot.cs
```diff
@@ -22,7 +22,7 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
             NpgsqlModelBuilderExtensions.UseIdentityByDefaultColumns(modelBuilder);
 
-            modelBuilder.Entity("API.ContractService.CertificateIssuingContract", b =>
+            modelBuilder.Entity("DataContext.Models.CertificateIssuingContract", b =>
```
tnickelsen, line 25: Did you test that this works when using dbupdate?

tnickelsen (approved):

domains/certificates/Shared/DataContext/Migrations/ApplicationDbContextModelSnapshot.cs
duizer: I have tested that it works when adding new migrations. Also, the integration tests apply the migrations

domains/certificates/README.md
duizer: You can do that, but you need to generate both Worker.sql and API.sql. Since the content of the .sql files are the same, it is faster to generate the .sql file once and then copy it.

tnickelsen: Why do we need both?

duizer: The build Github Action pipeline expects a .sql-file for each application/service in the sub-system, so it can verify that the .sql-file has been correctly updated. 
If only a single file should be the result, then we need to change the Github Action pipeline and add e.g. a configuration file instructing the pipeline how to verify the SQL script(s). After talking with Jakob we found the amount of work needed for that, is not worth it.

PR #554, opened 2023-12-28
Preview/claim automation refactor 1792

ckr123:


domains/transfer/Shared/DataContext/Migrations/20230515102933_InitialCreate.cs
```diff
@@ -1,7 +1,7 @@
 using System;
 using Microsoft.EntityFrameworkCore.Migrations;
 
-namespace API.Shared.Migrations;
+namespace DataContext.Migrations;
```
tnickelsen, line 4: Did you test that new migrations are added to this folder?

ckr123: Updated the readme in 20246f77 to support the new migration structure

tnickelsen (approved):

PR #571, opened 2024-01-15
Update certificates nugets

sahma19:


domains/certificates/Query.API/API.IntegrationTests/UnhealthTests.cs
```diff
@@ -2,18 +2,23 @@
...
-public class UnhealthTests : TestBase, IClassFixture<QueryApiWebApplicationFactory>
+public class UnhealthTests :
+    TestBase,
+    IClassFixture<QueryApiWebApplicationFactory>,
+    IClassFixture<PostgresContainer>
 {
     private readonly QueryApiWebApplicationFactory factory;
 
-    public UnhealthTests(QueryApiWebApplicationFactory factory)
+    public UnhealthTests(QueryApiWebApplicationFactory factory, PostgresContainer dbContainer)
```
tnickelsen, line 17: Why this change?

domains/certificates/Query.API/API/Program.cs
```diff
@@ -52,7 +52,8 @@
 builder.Services.AddDbContextFactory<ApplicationDbContext>();
 
 builder.Services.AddHealthChecks()
-    .AddNpgSql(builder.Configuration.GetConnectionString("Postgres")!);
+    .AddNpgSql(sp => sp.GetRequiredService<IConfiguration>().GetConnectionString("Postgres")!);
```
tnickelsen, line 55: Why this change?

domains/certificates/RegistryConnector/Worker/Program.cs
```diff
@@ -89,11 +89,11 @@
 
         cfg.ConfigureEndpoints(context);
 
-        cfg.ConfigureJsonSerializerOptions(options =>
+        cfg.ConfigureJsonSerializerOptions(jsonSerializerOptions =>
```
tnickelsen, line 92: Why change the name?

domains/certificates/RegistryConnector/Worker/Program.cs
```diff
@@ -54,7 +54,7 @@
 builder.Services.AddScoped<IKeyGenerator, KeyGenerator>();
 
 builder.Services.AddHealthChecks()
-    .AddNpgSql(builder.Configuration.GetConnectionString("Postgres")!);
+    .AddNpgSql(sp => sp.GetRequiredService<IConfiguration>().GetConnectionString("Postgres")!);
```
tnickelsen, line 57: Why this change?

domains/certificates/RegistryConnector/Worker/ProjectOriginOptions.cs
```diff
@@ -2,6 +2,7 @@
 using System.ComponentModel.DataAnnotations;
 using System.Text;
 using Microsoft.Extensions.DependencyInjection;
+using ProjectOrigin.HierarchicalDeterministicKeys.Implementations;
```
tnickelsen, line 5: This can be deleted?

domains/certificates/RegistryConnector/Worker/Program.cs
sahma19: After upgrading a healthcheck nuget package, the old way does not work anymore, causing tests to fail. Therefore the serviceprovider is now used to register the npgsql health endpoint.

domains/certificates/RegistryConnector/Worker/ProjectOriginOptions.cs
sahma19: I don't think so. I upgraded project origin, and then it wanted this dependency

domains/certificates/Query.API/API/Program.cs
sahma19: After upgrading a healthcheck nuget package, the old way does not work anymore, causing tests to fail. Therefore the serviceprovider is now used to register the npgsql health endpoint.

domains/certificates/Query.API/API.IntegrationTests/UnhealthTests.cs
sahma19: This test checks if rabbitmq, returns 503, if it is not available, but due to the fact that rabbitmq uses postgresql, for some outerbox, it requires postgres to be running before it can be started, before we can check rabbitmq's health, so i added a postgresdb container to be spun up.

tnickelsen (approved):

PR #575, opened 2024-01-17
Update auth nugets

sahma19:
This PR is part of the NuGet packages upgrade, happening across all subsystems in EO:

https://app.zenhub.com/workspaces/team-atlas-633199659e255a37cd1d144f/issues/gh/energinet-datahub/energy-origin-issues/1868


domains/auth/Auth.API/Unit.Tests/SwaggerTests.cs
```diff
@@ -2,7 +2,6 @@
 
 namespace Unit.Tests;
 
-[UsesVerify]
```
tnickelsen, line 5: Why has this been removed?

sahma19: Newest version of Verify, does not use this anymore. It is implicitly enabled by default, and can be invoked by just calling the api in the test

tnickelsen (approved):

PR #580, opened 2024-01-22
Added migration for new MassTransit package version

tnickelsen:


sahma19 (approved):
reluctantly approving ;)

PR #591, opened 2024-01-25
Add tests and update api package

TopSwagCode:


martinhenningjensen (dismissed):

libraries/dotnet/EnergyOrigin.ActivityLog/EnergyOrigin.ActivityLog/ActivityLogExtensions.cs
```diff
@@ -20,7 +20,7 @@ public static IServiceCollection AddActivityLog(this IServiceCollection services
     public static RouteHandlerBuilder UseActivityLog(this IEndpointRouteBuilder builder)
     {
         return builder.MapPost(
-            "/activity-log",
+            "api/activity-log",
```
sahma19, line 23: Explain to me what this does. Is the library itself hosting an endpoint now?

And is that endpoint "api/activity"?

Or is that endpoint "api/{name-of-entity}/activity-log"?

sahma19:
@TopSwagCode @martinhenningjensen

Are you both contributors to the overall issue, or is it only @TopSwagCode?

I believe having other teammates, than yourselves review and approve the code, would be more beneficial. It makes sure that other teammembers are kept in the loop.

sahma19 (changes requested):

libraries/dotnet/EnergyOrigin.ActivityLog/EnergyOrigin.ActivityLog/ActivityLogExtensions.cs
TopSwagCode: The package is not hosting an API endpoint, but adding support for the other API's can host it.

So the endpoint is: "api/activity-log" <-- activity-log is name of the resource. Just forgot to add api part to follow other API endpoints.

TopSwagCode: The first version of nuget package was merged because we wanted a build of the nuget package, so we could work on implementing logic in the different submodules / sub systems, while doing final work on nuget package.

TopSwagCode:
> @TopSwagCode @martinhenningjensen
> 
> Are you both contributors to the overall issue, or is it only @TopSwagCode?
> 
> I believe having other teammates, than yourselves review and approve the code, would be more beneficial. It makes sure that other teammembers are kept in the loop.

Agree :) Not meant to be only Martin approving. Just so we both look at each other code in the process. But in this current PR it's only be who has been doing some work on it, while Martin has been working on other parts in other PR.

sahma19 (approved):

PR #664, opened 2024-02-16
Changed entityId to be a string

tnickelsen:


libraries/dotnet/EnergyOrigin.ActivityLog/EnergyOrigin.ActivityLog/API/ActivityLogEntryResponse.cs
```diff
@@ -22,5 +22,5 @@ public enum ActionTypeEnum { Created, Accepted, Declined, Activated, Deactivated
     // Action
     public EntityTypeEnum EntityType { get; init; }
     public ActionTypeEnum ActionType { get; init; }
-    public Guid EntityId { get; init; }
+    public string EntityId { get; init; } = "";
```
sahma19, line 25: Why should it be a string?

sahma19 (changes requested):

libraries/dotnet/EnergyOrigin.ActivityLog/EnergyOrigin.ActivityLog/API/ActivityLogEntryResponse.cs
tnickelsen: Because we want to display GSRN number and not contract ID fx

sahma19 (approved):

PR #682, opened 2024-02-23
2444 Third party mock for auth flow

tnickelsen:


domains/third-party-mock/Ralunarg/BackgroundServices/SampledBackgroundService.cs
```diff
@@ -0,0 +1,31 @@
...
+    private readonly TokenClient _tokenClient;
+    private readonly TeamsClient _teamClient;
+
+    public SampledBackgroundService(TokenClient tokenClient, TeamsClient teamClient)
+    {
+        _tokenClient = tokenClient;
+        _teamClient = teamClient;
+    }
+
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
```
TopSwagCode, line 20: I would perhaps move this into a Console application for Client Credentials flow for getting access token and posting in channel. So we can run it as we see fit, instead of waiting for 1 minute intervals.

PR #783, opened 2024-03-07
Vcluster/2361 shared nuget for setting up EnergyOrigin solutions

tnickelsen:


libraries/dotnet/EnergyOrigin.Setup/EnergyOrigin.Setup/Swagger/ConfigureSwaggerOptions.cs
```diff
@@ -0,0 +1,59 @@
...
+                    },
+                    new string[] { }
+                }
+            });
+
+        foreach (var description in provider.ApiVersionDescriptions)
+        {
+            options.SwaggerDoc(description.GroupName, new OpenApiInfo
+            {
+                Title = $"{subsystemName.FirstCharToUpper()} Query API {description.ApiVersion}",
```
ckr123, line 53: Could this title be a variable?

libraries/dotnet/EnergyOrigin.Setup/EnergyOrigin.Setup/EnergyOrigin.Setup.csproj
```diff
@@ -0,0 +1,37 @@
...
+	</PropertyGroup>
+
+	<ItemGroup>
+		<FrameworkReference Include="Microsoft.AspNetCore.App" />
+	</ItemGroup>
+
+	<ItemGroup>
+	  <PackageReference Include="Asp.Versioning.Mvc" Version="8.0.0" />
+	  <PackageReference Include="Asp.Versioning.Mvc.ApiExplorer" Version="8.0.0" />
+	  <PackageReference Include="MassTransit" Version="8.1.3" />
```
ckr123, line 23: I dont believe all our systems are using masstransit. Do we still want it here or could we make some kind of builder pattern on the metrics/traces so it is optional

martinhenningjensen (approved):

libraries/dotnet/EnergyOrigin.Setup/EnergyOrigin.Setup/Swagger/ConfigureSwaggerOptions.cs
tnickelsen: removed query from the string

libraries/dotnet/EnergyOrigin.Setup/EnergyOrigin.Setup/EnergyOrigin.Setup.csproj
tnickelsen: We keep it to have masstransit tracing on on default. They will just be empty if there's no masstransit in the project

libraries/dotnet/EnergyOrigin.Setup/EnergyOrigin.Setup/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+            .AddApiExplorer();
+    }
+
+    public static void AddControllersWithEnumsAsStrings(this IServiceCollection services)
+    {
+        services.AddControllers()
+            .AddJsonOptions(o =>
+                o.JsonSerializerOptions.Converters.Add(new JsonStringEnumConverter()));
+    }
+}
```
sahma19, line 37: If this is added to the NuGet, are we then forcing everyone who imports it, to use versioning in our api?

I suggest having a boolean as a featureflag, that makes one able toggle on/off, each feature that is included in the NuGet. Be it MassTransit, Swagger, JsonEnumStringConverter, etc.

sahma19 (changes requested):

libraries/dotnet/EnergyOrigin.Setup/EnergyOrigin.Setup/IServiceCollectionExtensions.cs
tnickelsen: No only if you use the specified function for versioning. I would rather make a new function for without versioning, if that is needed, than making a boolean. But do we have apis that does not need versioning?

ckr123 (approved):

sahma19 (approved):

PR #784, opened 2024-03-08
Update validation of dotnet projects

CodeReaper:
- Do not run tests for already tested and approved changes that are merged in main.
- Output logs for failing tests and a single passed line for passing tests.
- Use version 4 of setup-dotnet action.


ole-kaas (approved):

sahma19 (changes requested):
So how exactly would this work?

Say I open up a pull request, there is new business logic, and 10 unit tests for it.

The pipeline will run the validation step, and it will pass if all 10 unit tests pass.

Let's say I afterwards make a new commit to the pull request. A commit that slightly alters the business logic.

Would the pipeline still run the 10 unit tests again?

CodeReaper:
> So how exactly would this work?
> 
> Say I open up a pull request, there is new business logic, and 10 unit tests for it.
> 
> The pipeline will run the validation step, and it will pass if all 10 unit tests pass.
> 
> Let's say I afterwards make a new commit to the pull request. A commit that slightly alters the business logic.
> 
> Would the pipeline still run the 10 unit tests again?

Yes - This only changes what happens after a PR is merged.

sahma19 (approved):

PR #797, opened 2024-03-13
TransferAgreementCleanup now deletes TA history as well

tnickelsen:


domains/transfer/Transfer.API/API/Transfer/TransferAgreementCleanup/TransferAgreementCleanupWorker.cs
```diff
@@ -46,6 +46,11 @@ private async Task DeleteExpiredTransferAgreements(CancellationToken cancellatio
         var expiredTransferAgreements = context.TransferAgreements
             .Where(ta => ta.EndDate != null && ta.EndDate < DateTimeOffset.UtcNow);
 
+        var expiredTaIds = expiredTransferAgreements.Select(ta => ta.Id).ToList();
+        var historyEntries = context.TransferAgreementHistoryEntries
+            .Where(h => expiredTaIds.Contains(h.TransferAgreementId));
```
sahma19, line 51: So what will happen if a TA is deleted? Will the foreign key be set to null in the history entry or will it be deleted as well?

martinhenningjensen (approved):

domains/transfer/Transfer.API/API/Transfer/TransferAgreementCleanup/TransferAgreementCleanupWorker.cs
tnickelsen: We're deleting both here. If I only deleted the TA then the FK in histories would remain the same and give an error - which is what we got

sahma19: Why not set the foreign key to null, and leave it at that? Shouldn't history be preserved?

sahma19: Same question for activity log

tnickelsen: You have no way of seeing it in the UI anymore as you can't click the TA. We are going towards using activity log and not history. The activity log should be preserved - yes. History is something to be deleted

PR #830, opened 2024-03-15
2361 measurements uses shared nuget for setup

tnickelsen:


domains/measurements/Measurements.API/API/Startup.cs
```diff
@@ -52,37 +39,15 @@ public void ConfigureServices(IServiceCollection services)
 
         services.AddEndpointsApiExplorer();
 
-        services.AddTransient<IConfigureOptions<SwaggerGenOptions>, ConfigureSwaggerOptions>();
+        services.AddSwagger("measurements");
         services.AddSwaggerGen();
```
ckr123, line 43: Shouldnt this be part of the nuget?

tnickelsen: No because that is used to register dokument filters (which is where have our api descriptions)

ckr123 (approved):

PR #869, opened 2024-03-20
2544 omit activitylog from swagger

tnickelsen:


domains/transfer/Shared/DataContext/DataContext.csproj
```diff
@@ -8,7 +8,7 @@
 
   <ItemGroup>
     <PackageReference Include="Audit.EntityFramework.Core" Version="25.0.1" />
-    <PackageReference Include="EnergyOrigin.ActivityLog" Version="4.0.0" />
+    <PackageReference Include="EnergyOrigin.ActivityLog" Version="4.0.2" />
```
sahma19, line 11: Why is it removed from transfer? Transfer api is not published to the developer portal. Only certificates api is

tnickelsen: Good point. I talked with Christoffer and we agreed to remove it nevertheless so we wouldn't have to make the nuget configureable

ckr123 (approved):

PR #911, opened 2024-03-27
2587 certificates unitofwork

tnickelsen:


TopSwagCode (approved):

domains/certificates/Query.API/API/ContractService/ContractServiceImpl.cs
```diff
@@ -70,7 +74,20 @@ public async Task<CreateContractResult> Create(string gsrn, string meteringPoint
 
         try
         {
-            await repository.Save(contract);
+            await unitOfWork.CertificateIssuingContractRepo.Save(contract);
+            await unitOfWork.ActivityLogEntryRepo.AddActivityLogEntryAsync(ActivityLogEntry.Create(
```
TopSwagCode, line 78: await unitOfWork.ActivityLogEntryRepo.AddActivityLogEntryAsync(ActivityLogEntry.Create(

Shouldn't this be part of UOW aswell???

TopSwagCode: We would need transaction scope to ensure that ActivityLog is only saved if whole transaction goes fine.

tnickelsen: It is as I am using the repo from the unit of work?

Why? SaveChanges is called to ensure this

TopSwagCode: await unitOfWork.ActivityLogEntryRepo.AddActivityLogEntryAsync(ActivityLogEntry.Create()) <-- just don't understand why this is Async then?

TopSwagCode (approved):

PR #953, opened 2024-04-09
add support for missing quantity measurements

martinhenningjensen:
Extend MeasurementSyncerWorker to support measurements marked with missing quantity. Info about missing measurements is persisted and data fetch is retried later.


domains/certificates/Query.API/API.UnitTests/MeasurementsSyncer/SlidingWindowServiceTest.cs
```diff
@@ -0,0 +1,337 @@
...
+        var fetchIntervalStart = window.GetFetchIntervalStart();
+
+        // Assert interval starts at earliest missing interval
+        Assert.Equal(now.Add(TimeSpan.FromHours(-6)), fetchIntervalStart);
+    }
+
+    [Theory]
+    [InlineData(false, 2)]
+    [InlineData(true, 0)]
+    public void GivenMeasurements_WhenPublishing_NewMeasurementsShouldBePublished(bool quantityMissing, int publishedCount)
```
tnickelsen, line 70: What is quantityMissing parameter used for?

domains/certificates/Shared/DataContext/ApplicationDbContext.cs
```diff
@@ -53,3 +66,5 @@ public ApplicationDbContext CreateDbContext(string[] args)
         return new ApplicationDbContext(optionsBuilder.Options);
     }
 }
+
+public class UnixTimestampConverter() : ValueConverter<UnixTimestamp, long>(v => v.Seconds, v => UnixTimestamp.Create(v));
```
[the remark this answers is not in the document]
martinhenningjensen: It's applied in ApplicationDbContext::ConfigureConventions method, to map the value object to and from an integer column in the database.

domains/certificates/Query.API/API.UnitTests/MeasurementsSyncer/SlidingWindowServiceTest.cs
martinhenningjensen: Test if quantityMissing is false, then 2 measurements are published. If quantityMissing is true, then no measurements are published.

martinhenningjensen: Yes, so the parameter is used in the test, when creating dummy measurements.

tnickelsen: In the test example, how can there be a quantity missing in a 1 hour window? I know it's just a test but if it was realistic. Maybe we can talk on teams about it

PR #954, opened 2024-04-09
remove history

ckr123:


domains/transfer/Shared/DataContext/Migrations/20240409063522_RemoveTransferAgreementHistory.cs
```diff
@@ -0,0 +1,50 @@
...
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DropTable(
+                name: "TransferAgreementHistoryEntries");
+        }
+
+        /// <inheritdoc />
+        protected override void Down(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.CreateTable(
```
TopSwagCode, line 21: Don't think we should have a down function here. If we have dropped the table, the data is gone forever.

domains/transfer/Transfer.API/API.IntegrationTests/DbContextExtension.cs
```diff
@@ -34,12 +34,9 @@ public static async Task<List<T>> RepeatedlyQueryUntilCountIsMet<T>(this DbConte
     public static async Task TruncateTransferAgreementsTables(this ApplicationDbContext dbContext)
```
TopSwagCode, line 34: We should delete this entire method. If we have dropped the table, we don't need to truncate it no more.

TopSwagCode (approved):

TopSwagCode (dismissed):

TopSwagCode (approved):

PR #1015, opened 2024-04-16
add step to wait for migrations in issuing cleanup

martinhenningjensen:


domains/certificates/Query.API/API/IssuingContractCleanup/IssuingContractCleanupWorker.cs
```diff
@@ -25,7 +25,14 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
 
             try
             {
-                await DeleteExpiredIssuingContracts(stoppingToken);
+                if (await IsWaitingForMigrations(stoppingToken))
```
tnickelsen, line 28: Can we create 2 dbContexts after eachother?

martinhenningjensen: Yes, seems so, but now you mention this... this service lacks scoping inside the loop.

tnickelsen: Why do we need the scope inside the loop? Isn't injecting the factory enough?

martinhenningjensen: To keep it consistent with other worker classes

tnickelsen (approved):

PR #1055, opened 2024-04-19
add integration test collection

martinhenningjensen:
this commit adds an integration test collection that controls all test containers. The collection is shared between integration tests to avoid stopping and starting test containers again and again.


domains/certificates/Query.API/API.IntegrationTests/IssuingContractCleanup/IssuingContractCleanupTests.cs
```diff
@@ -88,4 +85,10 @@ public async Task ShouldOnlyDeleteExpiredIssuingContracts()
         contracts.Select(x => x.Id).Should().Contain(nullEndDateContract.Id);
         contracts.Select(x => x.Id).Should().Contain(endDateContract.Id);
     }
+
+    private ApplicationDbContext GetDbContext()
+    {
+        var contextOptions = new DbContextOptionsBuilder<ApplicationDbContext>().UseNpgsql(postgresContainer.ConnectionString).Options;
+        return new ApplicationDbContext(contextOptions);
```
ckr123, line 92: Wont we benefit of setting the using statement here so we dont forget it? eg. line `28` :)

martinhenningjensen: Nope, because the using statement will dispose the object when it goes out of scope. The scope in this case is the method, however the object is returned from the method. It is therefore the responsibility of the caller to dispose the object.

ckr123 (approved):

ckr123 (approved):

PR #1105, opened 2024-04-29
fix conflict multiple update

ckr123:


domains/certificates/Query.API/API/ContractService/ContractServiceImpl.cs
```diff
@@ -60,8 +61,11 @@ internal class ContractServiceImpl : IContractService
                 return new GsrnNotFound();
             }
 
-            var overlappingContract = contractsByGsrn.FirstOrDefault(c =>
+            var contractsGsrn = contractsByAllGsrn.Where(c => c.GSRN == contract.GSRN).ToList();
```
martinhenningjensen, line 64: What happens if the request contains multiple contracts on the same GSRN number? 

It looks like contractsByAllGsrn variable is not updated with contracts created as part of the request handling?

domains/certificates/Query.API/API/ContractService/ContractServiceImpl.cs
```diff
@@ -161,7 +164,6 @@ private static MeteringPointType Map(MeterType type)
         var contractsByGsrn =
             await unitOfWork.CertificateIssuingContractRepo.GetByGsrn(issuingContracts.Select(c => c.GSRN).ToList(),
                 cancellationToken);
-
         foreach (var updatedContract in contracts.Contracts)
```
martinhenningjensen, line 167: Does this have the same issue as mentioned previously?

martinhenningjensen (changes requested):

domains/certificates/Query.API/API/ContractService/ContractServiceImpl.cs
ckr123: fixed in 1eb3207b

domains/certificates/Query.API/API/ContractService/ContractServiceImpl.cs
ckr123: fixed in 1eb3207b

martinhenningjensen: I don't see a test where multiple contracts with same GSRN are bulk-created :)

martinhenningjensen (approved):

PR #1132, opened 2024-05-02
Fix Testing.csproj setup

sahma19:
Having XUnit as a dependency inside a project flags it as a testing project for the dotnet builder. 

Removed Xunit, FluentAssertions, and NSubstitute from the Testing.csproj file.

Disposal of resources should be handled by the projects utilizing the testcontainers, and not be implemented in the test container through xunits IAsyncLifetime


domains/transfer/Testing/Testcontainers/RabbitMqContainer.cs
```diff
@@ -1,11 +1,10 @@
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Testcontainers.RabbitMq;
-using Xunit;
 
 namespace Testing.Testcontainers;
 
-public partial class RabbitMqContainer : IAsyncLifetime
+public partial class RabbitMqContainer
```
martinhenningjensen, line 7: Not used anywhere in the solution, this can be deleted.

domains/transfer/Testing/Testcontainers/PostgresContainer.cs
```diff
@@ -1,10 +1,9 @@
 using System.Threading.Tasks;
 using Testcontainers.PostgreSql;
-using Xunit;
 
 namespace Testing.Testcontainers;
 
-public class PostgresContainer : IAsyncLifetime
+public class PostgresContainer
```
martinhenningjensen, line 6: This is used as test fixture, I believe this should implement IAsyncLifetime in order to clean up resources.

martinhenningjensen (changes requested):

domains/transfer/Testing/Testcontainers/PostgresContainer.cs
sahma19: That would add back xUnit as dependency. Then we are back where we started

PR #1145, opened 2024-05-09
Transfer workflow

sahma19:


.github/workflows/build-subsystems.yaml
```diff
@@ -0,0 +1,114 @@
...
+    steps:
+      - name: detect auth
+        id: auth
+        uses: CodeReaper/find-diff-action@v3
+        with:
+          paths: domains/auth/
+
+      - name: detect authorization
+        id: authorization
+        uses: CodeReaper/find-diff-action@v3
```
martinhenningjensen, line 34: Should use an action in acorn repo instead

doc/ci-cd/workflows/specific-subsystem-workflows.md
```diff
@@ -0,0 +1,123 @@
+﻿## Specific Workflows for Each Subsystem
```
martinhenningjensen, line 1: There seems to be a lot of duplicate description from readme.me in this :)

.github/workflows/build-subsystems.yaml
```diff
@@ -0,0 +1,114 @@
...
+        description: "An indication of whether to commit/publish results"
+        required: true
+        type: string
+      is-dependabot:
+        description: "An indication of a dependabot pull request"
+        required: true
+        type: string
+
+jobs:
+  setup:
```
martinhenningjensen, line 16: I don't like this, but now at least we know where to change it if/when the assumptions no longer hold :)

martinhenningjensen (changes requested):

.github/workflows/build-subsystems.yaml
sahma19: Yes. or fork the action to our own repository. both are viable

doc/ci-cd/workflows/specific-subsystem-workflows.md
sahma19: I removed it

domains/transfer/change.txt
```diff
@@ -0,0 +1 @@
+trigger stuff
```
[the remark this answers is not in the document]
sahma19: In order to trigger a workflow we need to make a change in a subsystem. so we added a dummy file

.github/actions/dotnet-build/action.yaml
```diff
@@ -0,0 +1,144 @@
...
+    - name: Fail - if versions are not present
+      shell: bash
+      run: |
+        set +e
+        ! test -z "${{ env.SDK }}"
+        ! test -z "${{ env.RUNTIME }}"
+
+    - name: Fetch Dockerfile
+      shell: bash
+      run: wget https://raw.githubusercontent.com/Energinet-DataHub/acorn-actions/adjust-dotnet-validations/Dockerfile.simplified -O Dockerfile # FIXME: this
```
martinhenningjensen, line 62: Should this be fetched from a branch in Acoun repo? Why not use the dockerfile in our own repo?

martinhenningjensen (changes requested):

.github/workflows/build-auth.yaml
```diff
@@ -0,0 +1,54 @@
...
+      is-dependabot:
+        description: "An indication of a dependabot pull request"
+        required: true
+        type: string
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
martinhenningjensen, line 19: This is used a lot, is it really necessary?
I would expect a single checkout the begin with.

martinhenningjensen (dismissed):

.github/workflows/build-auth.yaml
sahma19: It is necessary. each job is running in its own environment. they all need to check out the code, because they boot up, with no files.

.github/actions/dotnet-build/action.yaml
sahma19: The platform requires that we release our deployments as docker images. it requires that they follow specific practices. Ratatosk has agreed to handle the responsibility of the Dockerfile, because of that. That is a win in my book.

.github/workflows/build-subsystems.yaml
sahma19: I don't understand

.github/workflows/build-subsystems.yaml
sahma19: The action now exists locally. Documentation for the action has been added as well.

Link to commit: https://github.com/Energinet-DataHub/energy-origin/pull/1145/commits/5cb6dd4d44ff6345f6ec499baaf447fbba09c73e

.github/actions/find-diff-action/action.yaml
```diff
@@ -0,0 +1,112 @@
...
+  using: composite
+  steps:
+    - name: Find diff
+      id: diff
+      shell: bash
+      env:
+        PATHS: ${{ inputs.paths }}
+        GH_TOKEN: ${{ github.token }}
+      run: |
+        echo "Configuration:"
```
TopSwagCode, line 35: Ohhh my bash :S

doc/workflows/find-diff-action.md
```diff
@@ -0,0 +1,83 @@
...
+By default all changes are reported, but the action can be configured to only consider certain folders or files and certain depths.
+
+The action produces both a list of relevant changes and a pattern for these changes.
+
+# Usage
+
+## List all changes
+
+```yaml
+- uses: CodeReaper/find-diff-action@v3
```
TopSwagCode, line 14: Seems like we are not using this anymore? Isn't diff inside here now?

sahma19: Fixed: https://github.com/Energinet-DataHub/energy-origin/pull/1145/commits/6679331224632b3ef0dbabfab8bee5f2ea9b314c

.github/actions/find-diff-action/action.yaml
sahma19: Et nødvendigt onde

TopSwagCode (approved):

martinhenningjensen (approved):

PR #1147, opened 2024-05-11
Vcluster/sigstore and gh

sahma19:


.github/workflows/build-domain.yaml
TopSwagCode, line 1: Why delete stuff when we need to be support old workflows?

.github/workflows/build-subsystems.yaml
```diff
@@ -0,0 +1,67 @@
...
+        subsystem: ${{ fromJson(needs.determine-subsystems.outputs.subsystems-json) }}
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+
+      - name: Dynamically Invoke Subsystem Workflow
+        uses: ./.github/actions/dynamic-uses
+        with:
+          uses: ./.github/actions/${{ matrix.subsystem }}
```
TopSwagCode, line 56: Why all this dynamic logic when we agreed to keep it simple?

domains/authorization/Authorization.API/API/Program.cs
```diff
@@ -61,3 +61,4 @@ public partial class Program
     {
     }
 }
+// Path:
```
TopSwagCode, line 64: ?

domains/transfer/Transfer.API/API/Program.cs
```diff
@@ -87,3 +87,4 @@
 public partial class Program
 {
 }
+//trigger
```
TopSwagCode, line 90: ?

TopSwagCode:
I don't really get this. We talked about keeping it all simple. But this seems to be following a lot of stuff that the old pipeline also did.

.github/workflows/build-domain.yaml
TopSwagCode: 1.  Make pipeline for transfer in new branch
2.  The pipeline must do: donet format, dotnet test, publish docker
3.  docker image is published side by side with existing without being used

PR #1153, opened 2024-05-14
add new b2c authentication/authorization

martinhenningjensen:


libraries/dotnet/EnergyOrigin.TokenValidation/EnergyOrigin.TokenValidation/B2C/B2COptions.cs
```diff
@@ -0,0 +1,26 @@
+using System.ComponentModel.DataAnnotations;
+
+namespace EnergyOrigin.TokenValidation.b2c;
+
+public class B2COptions
+{
+    public const string Prefix = "B2C";
+
+    [Required]
+    public string B2CWellKnownUrl { get; init; } = null!;
```
TopSwagCode, line 10: Maybe hardcode defaults for now, so we can test quicker? so we have other issue to create them in EO-BASE?

TopSwagCode (approved):

libraries/dotnet/EnergyOrigin.TokenValidation/EnergyOrigin.TokenValidation/B2C/B2COptions.cs
martinhenningjensen: Don't we include appsettings files in deployment? (The subsystem using token validation has the configuration in appsettings)

sahma19: You will need to add environment variables to the deployment, in eo-base :)

sahma19 (changes requested):
ENV variable

sahma19 (approved):

PR #1176, opened 2024-05-23
Quorum queues set

tnickelsen:


TopSwagCode (approved):

TopSwagCode:
Is there any reason for this change?

domains/certificates/RegistryConnector/Worker/Program.cs
```diff
@@ -69,6 +69,7 @@
         });
 
         cfg.ConfigureEndpoints(context);
+        cfg.SetQuorumQueue();
```
TopSwagCode, line 72: if(cfg is IRabbitMqEndpointConfigurator rmq)
        rmq.SetQuorumQueue();
        
Would recommend maybe that we make a check if it's rabbitmq. Otherwise it would break if we change to other queue in future.

TopSwagCode:
Can we delete this PR?

PR #1189, opened 2024-06-03
add b2c custom flow for oidcmock

martinhenningjensen:
Change org_ids claim to space seperated string og organization ids. Remove unused technical profiles in custom profiles. Remove objectId as output claim to fix flow.

### DIFF
--- a/libraries/dotnet/EnergyOrigin.TokenValidation/configuration.yaml
+++ b/libraries/dotnet/EnergyOrigin.TokenValidation/configuration.yaml
@@ -1,1 +1,1 @@
-version: 3.6.2
+version: 3.6.3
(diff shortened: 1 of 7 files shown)


TopSwagCode (dismissed):

libraries/dotnet/EnergyOrigin.TokenValidation/configuration.yaml
```diff
@@ -1 +1 @@
-version: 3.6.2
+version: 3.7.0
```
TopSwagCode, line 1: version bump? :D

martinhenningjensen: It's not just a patch version bump, but also not something that will break consumers with a compile error. How should the version be bumped?

TopSwagCode (approved):

PR #1240, opened 2024-06-30
Accept terms - appetizer?

sahma19:


domains/authorization/Authorization.API/API/Authorization/Controllers/TermsController.cs
```diff
@@ -0,0 +1,56 @@
...
+public class TermsController(IMediator mediator) : ControllerBase
+{
+    [ProducesResponseType(typeof(TermsResponseDto), StatusCodes.Status200OK)]
+    [HttpPost("accept")]
+    public async Task<IActionResult> AcceptTerms([FromBody] AcceptTermsDto acceptTermsDto)
+    {
+        var isValid = await mediator.Send(new OrganizationStateQuery(acceptTermsDto.Tin));
+
+        if (isValid)
+            return Ok(new TermsResponseDto(true));
```
TopSwagCode, line 24: TermsResponse is enough :) DTO is implied in response.

domains/authorization/Authorization.API/API/Authorization/Controllers/TermsController.cs
```diff
@@ -0,0 +1,56 @@
...
+[ApiController]
+[ApiVersion(ApiVersions.Version20230101)]
+[Route("api/authorization/terms")]
+public class TermsController(IMediator mediator) : ControllerBase
+{
+    [ProducesResponseType(typeof(TermsResponseDto), StatusCodes.Status200OK)]
+    [HttpPost("accept")]
+    public async Task<IActionResult> AcceptTerms([FromBody] AcceptTermsDto acceptTermsDto)
+    {
+        var isValid = await mediator.Send(new OrganizationStateQuery(acceptTermsDto.Tin));
```
TopSwagCode, line 21: Query could be something like: HasOrganizationAcceptedTermsQuery.

domains/authorization/Authorization.API/API/Authorization/Controllers/TermsController.cs
```diff
@@ -0,0 +1,56 @@
...
+namespace API.Authorization.Controllers;
+
+[ApiController]
+[ApiVersion(ApiVersions.Version20230101)]
+[Route("api/authorization/terms")]
+public class TermsController(IMediator mediator) : ControllerBase
+{
+    [ProducesResponseType(typeof(TermsResponseDto), StatusCodes.Status200OK)]
+    [HttpPost("accept")]
+    public async Task<IActionResult> AcceptTerms([FromBody] AcceptTermsDto acceptTermsDto)
```
TopSwagCode, line 19: Request body should be named xRequest, to imply its API request. Naming could be AcceptTermsRequest.

domains/authorization/Authorization.API/API/Authorization/Controllers/TermsController.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        var isValid = await mediator.Send(new OrganizationStateQuery(acceptTermsDto.Tin));
+
+        if (isValid)
+            return Ok(new TermsResponseDto(true));
+
+        var latestTerms = await mediator.Send(new GetLatestTermsQuery());
+
+        if (acceptTermsDto.TermsVersion != latestTerms.Version)
```
TopSwagCode, line 28: This logic should be in the query. To check if terms are latests terms.

domains/authorization/Authorization.API/API/Authorization/Controllers/TermsController.cs
```diff
@@ -0,0 +1,56 @@
...
+            return Ok(new TermsResponseDto(false, latestTerms.Text, latestTerms.Version));
+
+        var createResult = await mediator.Send(new CreateOrganizationAndUserCommand(
```
TopSwagCode, line 31: Nice and simple name :) Create Organization And User command :) States exactly what it does.

domains/authorization/Authorization.API/API/Authorization/Controllers/TermsController.cs
```diff
@@ -0,0 +1,56 @@
...
+            acceptTermsDto.UserName,
+            latestTerms.Version
+        ));
+
+        return Ok(new TermsResponseDto(true, CreateResult: createResult));
+    }
+}
+
+public record AcceptTermsDto(
+    string Tin,
```
TopSwagCode, line 44: All except accept terms can be taken from Claims. You shouldn't be able to accept on other organizations that the one your logged in as.

domains/authorization/Authorization.API/API/Authorization/Controllers/TermsController.cs
TopSwagCode, line 19: I would change this endpoint only to create organization and set terms version to whatever version we have. Not doing anything with user. Then let Get consent endpoint create / update user. So we don't have duplicate places that does same logic.

PR #1268, opened 2024-07-02
Accepted terms v2 added

tnickelsen:


sahma19 (approved):

TopSwagCode (approved):

libraries/dotnet/EnergyOrigin.IntegrationEvents/EnergyOrigin.IntegrationEvents/Events/IntegrationEvent.cs
```diff
@@ -0,0 +1,15 @@
+namespace EnergyOrigin.IntegrationEvents.Events;
+
+public abstract record IntegrationEvent
+{
+    protected IntegrationEvent(Guid id, Guid traceId, DateTimeOffset created)
+    {
+        Id = id;
```
ckr123, line 7: Why is Id needed when you have traceid?

tnickelsen: TraceId is for the trace. Id is so that we can uniquely identify the event. If we will ever need it, I dunno, but an architect decided this

PR #1269, opened 2024-07-02
add new policy, for acceptance of terms

sahma19:


libraries/dotnet/EnergyOrigin.TokenValidation/EnergyOrigin.TokenValidation/B2C/IServiceCollectionExtensions.cs
```diff
@@ -53,6 +53,15 @@ public static void AddB2CAndTokenValidation(this IServiceCollection services, B2
                 .Build();
             options.AddPolicy(Policy.B2CPolicy, b2CPolicy);
 
+            var b2CTosAcceptedPolicy = new AuthorizationPolicyBuilder()
+                .RequireAuthenticatedUser()
+                .AddAuthenticationSchemes(
+                    AuthenticationScheme.B2CClientCredentialsCustomPolicyAuthenticationScheme,
+                    AuthenticationScheme.B2CMitIDCustomPolicyAuthenticationScheme)
+                .RequireClaim(ClaimType.TosAccepted, "true")
```
TopSwagCode, line 61: This should just be part of already existing policies. This should be part of SubTypeUser.

TopSwagCode: Maybe this shouldn't be alone, since it will break everything until we have terms up and running.

sahma19: Making it a part of an existing policy, would result in a breaking change. this way, consumers of the nuget package, can opt-in, for the added functionality. 

I would rather have the functionality available like this, and when the feature is finished being implemented, we can talk about including it in the current policy.

TopSwagCode: Martin, You and me just had a talk about this issue? We agreed upon extending existing policy. No need to add a policy just to delete it afterwards again? The frontend is not using the new endpoint anyway, so not like we are blocking anyone. This will just add legacy code out of the gate.

TopSwagCode: Maybe this shouldn't be alone, since it will break everything until we have terms up and running. <-- This PR is not self contained. We can't use the policy as is. We need endpoints, logic, etc. We had this talk about the big issue being a 13 pointer and would take weeks.

TopSwagCode (changes requested):
We need to have working logic and not implement small things that wont work until we have more.

This would be one of the last steps. Because if anyone bumps the package we will have breaking functionality.

libraries/dotnet/EnergyOrigin.TokenValidation/EnergyOrigin.TokenValidation/B2C/IServiceCollectionExtensions.cs
sahma19: There is no breaking change, as long as the package is not updated in the subsystems

martinhenningjensen: Exactly, so we don't have the option of making other changes in this package before the rest of the functionality has been delivered. I agree that this part could be done last, when all functionality regarding the new claim and accepting terms have been delivered.

libraries/dotnet/EnergyOrigin.TokenValidation/EnergyOrigin.TokenValidation/B2C/TermsRequirementHandler.cs
```diff
@@ -0,0 +1,31 @@
+using EnergyOrigin.TokenValidation.b2c;
+using Microsoft.AspNetCore.Authorization;
+
+namespace EnergyOrigin.TokenValidation.B2C;
+
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Method)]
+public class DisableTermsRequirementAttribute : Attribute;
```
TopSwagCode, line 7: Why do we need a DisableTermsRequirementAttribute? If not terms policy is choosen, then it shouldn't be checked?

libraries/dotnet/EnergyOrigin.TokenValidation/doc/auth-subsystem/token-validation.md
```diff
@@ -0,0 +1,147 @@
...
+Token validation is built around facilitating the [Security and Identity](https://learn.microsoft.com/en-us/aspnet/core/security/authorization/introduction?view=aspnetcore-7.0) concept in ASP.NET Core.
+
+For our specific use cases, see [restricting access](#restricting-access).
+
+## Quick Start
+
+1. Register token validation during start up:
+
+```csharp
+var options = new TokenValidationOptions {
```
TopSwagCode, line 12: Why not just delete all Token Validation stuff, since it's not needed anymore.

libraries/dotnet/EnergyOrigin.TokenValidation/EnergyOrigin.TokenValidation/B2C/TermsRequirementHandler.cs
sahma19: Default behaviour should be to lock down all apis' from users unless terms are accepted. This way we don't have to write add a new terms policy in all controllers. we just disable the ones we want. Opt-out rather than opt-in.

libraries/dotnet/EnergyOrigin.TokenValidation/EnergyOrigin.TokenValidation/B2C/TermsRequirementHandler.cs
```diff
@@ -0,0 +1,31 @@
...
+        var endpoint = context.Resource as Microsoft.AspNetCore.Http.Endpoint;
+        var disableTermsAttribute = endpoint?.Metadata.GetMetadata<DisableTermsRequirementAttribute>();
+
+        if (disableTermsAttribute != null)
+        {
+            context.Succeed(requirement);
+            return Task.CompletedTask;
+        }
+
+        if (context.User.HasClaim(c => c is { Type: ClaimType.TosAccepted, Value: "true" }))
```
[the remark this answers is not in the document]
sahma19: When Old auth is gone, we can do so.

libraries/dotnet/EnergyOrigin.TokenValidation/doc/auth-subsystem/token-validation.md
sahma19: When Old auth is gone, we can do so.

libraries/dotnet/EnergyOrigin.TokenValidation/EnergyOrigin.TokenValidation/B2C/IServiceCollectionExtensions.cs
```diff
@@ -59,13 +60,15 @@ public static void AddB2CAndTokenValidation(this IServiceCollection services, B2
                     AuthenticationScheme.B2CClientCredentialsCustomPolicyAuthenticationScheme,
                     AuthenticationScheme.B2CMitIDCustomPolicyAuthenticationScheme)
                 .RequireClaim(ClaimType.SubType, Enum.GetName(SubjectType.User)!, Enum.GetName(SubjectType.User)!.ToLower())
+                .AddRequirements(new TermsRequirement())
```
TopSwagCode, line 63: 3rd party can never have accepted terms. They are not part of terms.

libraries/dotnet/EnergyOrigin.TokenValidation/EnergyOrigin.TokenValidation/B2C/TermsRequirementHandler.cs
```diff
@@ -0,0 +1,31 @@
...
+
+public class TermsRequirement : IAuthorizationRequirement;
+
+public class TermsRequirementHandler : AuthorizationHandler<TermsRequirement>
+{
+    protected override Task HandleRequirementAsync(AuthorizationHandlerContext context, TermsRequirement requirement)
+    {
+        var endpoint = context.Resource as Microsoft.AspNetCore.Http.Endpoint;
+        var disableTermsAttribute = endpoint?.Metadata.GetMetadata<DisableTermsRequirementAttribute>();
+
```
TopSwagCode, line 17: We need to check if it's 3rd party or not. If 3rd party they dont have terms accepted.

sahma19: That should not be relevant. The requirement for having accepted terms, is only applied for the policy for users. So unless someone uses a "user-policy" for authorizing clients, in their controller actions, then i don't see a problem

TopSwagCode:
What is state of this PR????

sahma19:
Solved in by @martinhenningjensen: https://github.com/Energinet-DataHub/energy-origin/pull/1414

Therefore closing this issue.

PR #1288, opened 2024-07-10
update certificates to support b2c downstream

martinhenningjensen:


domains/certificates/Query.API/API/ContractService/Clients/MeteringPointsClient.cs
```diff
@@ -33,22 +32,45 @@ public MeteringPointsClient(HttpClient httpClient, IHttpContextAccessor httpCont
...
 
     private void SetAuthorizationHeader()
+    {
+        httpClient.DefaultRequestHeaders.Authorization = AuthenticationHeaderValue.Parse(httpContextAccessor.HttpContext!.Request.Headers.Authorization!);
+    }
+
+    private void SetApiVersionHeader()
+    {
+        var downstreamApiVersion = GetDownstreamApiVersion();
+        httpClient.DefaultRequestHeaders.Remove("EO_API_VERSION");
```
sahma19, line 54: Why remove and add back?

martinhenningjensen: The add method adds an additional header with no check if a header with same name is already added. So we get requests with multiple EO_API_VERSION headers.

sahma19 (approved):

PR #1289, opened 2024-07-11
remove swagger test

ckr123:


sahma19:
The end of an era 😢

sahma19:
BTW. If you remove one of the swagger tests in certificates, you will break Redoc. You will have to find another way to publish the api spec as a release artifact for redoc to collect. Some guides here:

https://www.meziantou.net/generate-openapi-specification-at-build-time-from-the-code-in-asp-net-core.htm

https://gist.github.com/VincentH-Net/a950fb18363111ccac8e1fb64123647e

ckr123:
> BTW. If you remove one of the swagger tests in certificates, you will break Redoc. You will have to find another way to publish the api spec as a release artifact for redoc to collect. Some guides here:
> 
> https://www.meziantou.net/generate-openapi-specification-at-build-time-from-the-code-in-asp-net-core.htm
> 
> https://gist.github.com/VincentH-Net/a950fb18363111ccac8e1fb64123647e
> 
> I suggest you do the same, for wallet proxy. That way whenever a new swagger spec is published, redoc will know about it right away

Ja kunne godt se det gav nogle problemer. Det er også et Draft det her og tror den her opgave nok skal parkeres til efter vores spike vedr. dev portal

PR #1310, opened 2024-07-29
add default terms

sahma19:


domains/authorization/Authorization.API/API/Migrations/20240730083845_AddUniqueTermsVersionConstraint.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace API.Migrations
+{
+    /// <inheritdoc />
+    public partial class AddUniqueTermsVersionConstraint : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.CreateIndex(
+                name: "IX_Terms_Version",
```
ckr123, line 14: Why index on version?

domains/authorization/Authorization.API/API/Migrations/20240730104826_AddDefaultTerms.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace API.Migrations
+{
+    /// <inheritdoc />
+    public partial class AddDefaultTerms : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.InsertData(
```
ckr123, line 14: Why not use on migration file?

domains/authorization/Authorization.API/API/Migrations/20240730083845_AddUniqueTermsVersionConstraint.cs
sahma19: In order to enforce the unique constraint, it needs to look it up in the database, and creating an index for that, makes it more performant.

ckr123: But i dont believe you need to make it an index to get the unique constraint? I dont think we get that many versions that you will notice any performance issues

domains/authorization/Authorization.API/API/Migrations/20240730104826_AddDefaultTerms.cs
sahma19: Fixed: https://github.com/Energinet-DataHub/energy-origin/pull/1310/commits/99abc05de6b469ca9431a6920ed1dffb024aa2b3

ckr123 (approved):

PR #1315, opened 2024-07-31
Vcluster/test pipeline new

ckr123:


.github/workflows/build-openapi.yaml
```diff
@@ -0,0 +1,37 @@
+name: Run .NET Project with Swagger
```
sahma19, line 1: Why do we need a separate workflow for generating an openapi spec?

You could simply add the "--swagger" flag to the regular build workflow, so it is always included.

The way you do it now is building the same app twice. Once for deploying it, and another time to generate an openapi spec.

domains/certificates/Query.API/API/Program.cs
```diff
@@ -19,107 +22,126 @@
...
+            .HasApiVersion(ApiVersions.Version20240423AsInt)
+            .HasApiVersion(ApiVersions.Version20230101AsInt);
+        app.UseActivityLogWithB2CSupport().WithApiVersionSet(activityLogApiVersionSet)
+            .HasApiVersion(ApiVersions.Version20240515AsInt);
 
-app.Run();
 
-public partial class Program
-{
+        if (args.Contains("--swagger"))
```
sahma19, line 132: Is this code needed?

Swashbuckle+dotnet has built-in support for generating swagger spec on build. I think I showed you two links explaining how to do it, by modifying the ".csproj" file.

.github/workflows/build-openapi.yaml
```diff
@@ -0,0 +1,37 @@
...
+        uses: Energinet-DataHub/acorn-actions/actions/docker-image-version@v1
+
+      - name: Resolve image name
+        shell: bash
+        run: echo "name=$(yq '.name' 'domains/configuration.yaml')" | tee -a $GITHUB_ENV
+
+      - name: Build and push Docker image
+        uses: Energinet-Datahub/acorn-actions/actions/docker-build-and-push@v1
+        with:
+          dockerfile: ./domains/Dockerfile
```
sahma19, line 33: We should aim to use the dockerfile provided by ratatosk. It has been been designed for our needs, while being more performant and secure than the default offerings you can find online.

.github/workflows/build-subsystems.yaml
```diff
@@ -92,6 +100,7 @@ jobs:
       - build-measurements
       - build-oidc-mock
       - build-transfer
+      - build-openapi
```
sahma19, line 103: I don't think you need another workflow/job to build an openapi spec. Could we instead strive to incorporate it into the regular build workflows?

domains/authorization/Proxy/Program.cs
```diff
@@ -1,104 +1,120 @@
...
+        app.MapHealthChecks("/health");
+
+        app.UseHttpsRedirection();
+
+        app.UseAuthorization();
+
+        app.MapControllers();
+
+        app.MapSwagger();
+        if (args.Contains("--swagger"))
```
sahma19, line 105: Why wrap this in an "if-statement block"? And why hide it behind a "--swagger" annotation? make the code execute on every build. Make a unit test for it, to test if it works, and you are done.

domains/certificates/Query.API/API/Program.cs
```diff
@@ -19,107 +22,126 @@
...
+        {
+            var swaggerProvider = app.Services.GetRequiredService<ISwaggerProvider>();
+            var swagger = swaggerProvider.GetSwagger(ApiVersions.Version20240515);
+
+            File.WriteAllText(
+                Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)!, "contracts.yaml"),
+                swagger.SerializeAsYaml(OpenApiSpecVersion.OpenApi3_0)
```
sahma19, line 139: Redoc is able to convert json to yaml by itself. At least that is what it does in the init container. No need to do it here.

.github/workflows/build-subsystems.yaml
```diff
@@ -82,6 +82,14 @@ jobs:
       dry-run: ${{ inputs.dry-run }}
       is-dependabot: ${{ inputs.is-dependabot }}
 
+  build-openapi:
+    needs: test
+    uses: ./.github/workflows/build-openapi.yaml
+    secrets: inherit
+    with:
+      dry-run: ${{ inputs.dry-run }}
```
sahma19, line 90: You have dry run feature toggles enabled, for this. Should it have this?

domains/authorization/Proxy/Program.cs
```diff
@@ -1,104 +1,120 @@
...
-
-public partial class Program
-{
-
+        var builder = WebApplication.CreateBuilder(args);
+
+        // Add services to the container.
+
+        builder.Services.AddControllers();
+        // Learn more about configuring Swagger/OpenAPI at https://aka.ms/aspnetcore/swashbuckle
```
sahma19, line 24: Remove this comment.

sahma19 (changes requested):

domains/certificates/Query.API/API/Program.cs
ckr123: This is an easy way to handle multiple versions of the API

.github/workflows/build-openapi.yaml
ckr123: We only need to run dotnet run, no need to change their dockerfile to do it

.github/workflows/build-subsystems.yaml
ckr123: Its an easy way to get all files in one docker image

domains/certificates/Query.API/API/Program.cs
ckr123: the OpenApiDocument type makes you serialize as yaml or json, no need to serialize json and make it yaml in the cli

.github/workflows/build-subsystems.yaml
ckr123: I dont see why not

domains/authorization/Proxy/Program.cs
ckr123: removed

domains/authorization/Proxy/Program.cs
ckr123: This is not being run in the build process and we dont want to start our application

domains/authorization/Proxy/Program.cs
```diff
@@ -1,104 +1,118 @@
...
 
-var builder = WebApplication.CreateBuilder(args);
-
-// Add services to the container.
-
-builder.Services.AddControllers();
-// Learn more about configuring Swagger/OpenAPI at https://aka.ms/aspnetcore/swashbuckle
-builder.Services.AddEndpointsApiExplorer();
+public class Program
+{
```
TopSwagCode, line 16: minor comment - Not important. 
Do we need to wrap this in a public class Program???

ckr123: It is to get the args in the main function. Dno if there is another way?

TopSwagCode: https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/program-structure/main-command-line

TopSwagCode (dismissed):

PR #1359, opened 2024-08-18
Preview/seed for auth

sahma19:
- Demo database has been seeded with the personas, from old auth solution
  - Wiped Authorization database
  - Ran the SQL script, which is defined in this pull request
- Invalid users has been removed from relations table, in measurements
  - There were duplicate persona users, with the same Tin, but different Ids, causing the wiremock to not recognize them. These have been removed.

- Creating new preview environments that are hooked up to the demo oidc-mock policy will no longer send an "TermsAccepted" event, as the users are no longer created, but have been pre seeded.

### DIFF
--- a/domains/authorization/migrations/SEED.sql
+++ b/domains/authorization/migrations/SEED.sql
@@ -1,77 +1,69 @@
-BEGIN;
-
-CREATE OR REPLACE FUNCTION upsert_organization(
-    p_id UUID,
-    p_tin VARCHAR(20),
-    p_name VARCHAR(255),
-    p_terms_acceptance_date TIMESTAMPTZ,
-    p_terms_accepted BOOLEAN,
-    p_terms_version INTEGER
-) RETURNS VOID AS $$
+DO $$
+DECLARE
+organizations_empty boolean;
+    users_empty boolean;
+    affiliations_empty boolean;
+    terms_empty boolean;
 BEGIN
-INSERT INTO "Organizations" ("Id", "Tin", "Name", "TermsAcceptanceDate", "TermsAccepted", "TermsVersion")
-VALUES (p_id, p_tin, p_name, p_terms_acceptance_date, p_terms_accepted, p_terms_version)
-    ON CONFLICT ("Id") DO UPDATE SET
-    "Tin" = EXCLUDED."Tin",
-                              "Name" = EXCLUDED."Name",
-                              "TermsAcceptanceDate" = EXCLUDED."TermsAcceptanceDate",
-                              "TermsAccepted" = EXCLUDED."TermsAccepted",
-                              "TermsVersion" = EXCLUDED."TermsVersion";
-END;
-$$ LANGUAGE plpgsql;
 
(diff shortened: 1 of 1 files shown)


domains/authorization/migrations/SEED.sql
```diff
@@ -0,0 +1,69 @@
...
+            ('95d7be81-0cfb-4b52-9c92-33a45747fcef', '39293595', 'Company Inc.', CURRENT_TIMESTAMP, TRUE, 1),
+            ('b0918bf8-626d-4b35-a0ad-3f5beb6b6fef', '55555555', 'Fabrikant', CURRENT_TIMESTAMP, TRUE, 1),
+            ('cb1e0981-1ea7-404e-b189-260c58a46c68', '12345678', 'Startup I/S', CURRENT_TIMESTAMP, TRUE, 1);
+
+        RAISE NOTICE 'Organizations table seeded successfully.';
+ELSE
+        RAISE NOTICE 'Organizations table is not empty. Skipping seed data insertion.';
+END IF;
+
+SELECT COUNT(*) = 0 FROM "Users" INTO users_empty;
```
martinhenningjensen, line 36: This could be omitted

domains/authorization/migrations/SEED.sql
```diff
@@ -0,0 +1,69 @@
...
+            ('b3519236-aaf5-4eef-84b7-724dc8289fb2', '7f4b6c63-8fbd-4927-8202-c6731c00a0fc', 'Viggos Vindmøller'),
+            ('df7bfcf7-c2c8-442b-83cc-b46e6e8c8e72', 'b0918bf8-626d-4b35-a0ad-3f5beb6b6fef', 'Fabrikant'),
+            ('ec738dea-6903-490d-b012-9a2fe98406dc', '21e150bd-8a0b-4ba6-905b-7223db248af5', 'Bolighaj');
+
+        RAISE NOTICE 'Users table seeded successfully.';
+ELSE
+        RAISE NOTICE 'Users table is not empty. Skipping seed data insertion.';
+END IF;
+
+SELECT COUNT(*) = 0 FROM "Affiliations" INTO affiliations_empty;
```
martinhenningjensen, line 53: This could be omitted

domains/authorization/migrations/SEED.sql
```diff
@@ -0,0 +1,69 @@
...
+IF terms_empty THEN
+        INSERT INTO "Terms" ("Id", "Version")
+        VALUES ('0ccb0348-3179-4b96-9be0-dc7ab1541771', 1);
+
+        RAISE NOTICE 'Terms table seeded successfully.';
+ELSE
+        RAISE NOTICE 'Terms table is not empty. Skipping seed data insertion.';
+END IF;
+
+SELECT COUNT(*) = 0 FROM "Organizations" INTO organizations_empty;
```
martinhenningjensen, line 19: Could we do an upsert instead so we make sure organizations are updated on a redeploy?

martinhenningjensen (changes requested):

PR #1416, opened 2024-08-30
address security warnings from openapi image

sahma19:
Most of the security warnings come from the openapi image. this PR, addresses those issues.

Related PR, in eo-base:

[EO-BASE PR](https://github.com/Energinet-DataHub/eo-base-environment/pull/1284)

Before:

image

After:

image


domains/Dockerfile
```diff
@@ -1,13 +1,21 @@
-# Use the official .NET SDK image
-FROM mcr.microsoft.com/dotnet/sdk:8.0 AS build
+FROM mcr.microsoft.com/dotnet/sdk:8.0.401-noble AS build
```
TopSwagCode, line 1: why is sdk and runtime using different versions?

sahma19: The SDK does not matter here, but we need an sdk to build it. I simply upgraded it to the latest version. 

Remember, this image is "only" used to serve 2 swagger specs to eo-base. It actually does not matter if they match or not. What matters is the runtime they are shipped with. The previous one had over 100 vulnerabilities flagged. Those are now taken care of.

sahma19: Also latest sdk version is 8.0.401 and runtime version is 8.0.8 😉

With that logic they would never match

sahma19: Resolving this comment, as @TopSwagCode is unavailable

tnickelsen (approved):

PR #1456, opened 2024-09-09
Limit to security updates

TopSwagCode:


.github/dependabot.yml
```diff
@@ -17,8 +17,4 @@ updates:
     directory: /domains
     schedule:
       interval: daily
-    groups:
-      all:
-        applies-to: security-updates
-        patterns:
-          - "*"
```
tnickelsen, line 24: Hwo does removing this set it to security updates only?

sahma19 (approved):

PR #1546, opened 2024-10-06
Pipeline Perfection

sahma19:
This decreases our pipeline run time from ~9 minutes to under 4 minutes, by making different subsystems test and build in parallel, rather than sequentially. Also we are no longer forced to test/build every subsystem, if we only introduce changes to a single one, while still having one consolidated solution:

image

### DIFF
--- a/.github/workflows/build-authorization.yaml
+++ b/.github/workflows/build-authorization.yaml
@@ -14,6 +14,7 @@ on:
 
 jobs:
   test:
+    name: Test Authorization Subsystem
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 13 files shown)


sahma19:
3m49s: 
image

But we can do better :)

To be continued

sahma19:
Note to self on how to improve build time further:

- All tests currently output verbosely to stdout. By making only failing tests print a stacktrace, we could potentially increase pipeline speed further.

- Right now the check licenses workflow is implemented so that it must complete its run, before testing and building can begin, thus adding more overhead to the build time. This can be improved, by making the license check run in parallel with the build subsystems workflow, potentially shaving up to **29s** in build time.

PR #1568, opened 2024-10-17
Update acorn actions used

CodeReaper:
We are ignoring these nuget packages since we do not contain license information at least not any that conforms to the tool we are checking with:
- `jose-jwt`
- `Microsoft.AspNet.WebApi.Client`

These ignored licenses have *do* have licenses that are in the allowed licenses file. We just cannot automate checking them.

They should properly be checked for alternatives that can be validated, or perhaps make pull requests to them so they can be fixed.

Also note, that no nuget packages used in test projects are validated still - there is a TODO about re-enabling them.

### DIFF
--- a/.github/actions/dotnet-build/action.yaml
+++ b/.github/actions/dotnet-build/action.yaml
@@ -67,7 +67,7 @@ runs:
       run: wget https://raw.githubusercontent.com/Energinet-DataHub/acorn-actions/main/Dockerfile.simplified -O Dockerfile
 
     - name: Resolve image version
-      uses: Energinet-DataHub/acorn-actions/actions/docker-image-version@v2
+      uses: Energinet-DataHub/acorn-actions/actions/docker-image-version@v3
 
     - name: Resolve image name
       shell: bash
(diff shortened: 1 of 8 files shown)


CodeReaper:
@tnickelsen - While the check-license is disabled I cannot really know if this PR fixes everything, but I have run the health check to see if the updated actions work there.

tnickelsen:
> @tnickelsen - While the check-license is disabled I cannot really know if this PR fixes everything, but I have run the health check to see if the updated actions work there.

You may reenable them to test this :)

tnickelsen (dismissed):

CodeReaper:
Testing it lead to few changes...

The input for validating licenses is now the solution file, this however will also target test projects, but will also check everything in one go.

This means the packages that are ignored are now:

- jose-jwt
- MockQueryable.NSubstitute
- Microsoft.AspNet.WebApi.Client
- Testcontainers*
- origin/*
- xunit.categories

The `origin/*` entry are local dependencies.

tnickelsen (approved):

PR #1579, opened 2024-10-25
refactor Authorization API

martinhenningjensen:

PR #1583, opened 2024-11-04
Calling grant consent twice now results in conflict

tnickelsen:


martinhenningjensen:
Handled as HTTP 200 OK in PR https://github.com/Energinet-DataHub/energy-origin/pull/1579

PR #1584, opened 2024-11-04
introduce value objects in transfer

martinhenningjensen:


domains/libraries/EnergyOrigin.Domain/EnergyOrigin.Domain.ValueObjects.Tests/TinTest.cs
```diff
@@ -1,43 +1,43 @@
...
     [Fact]
     public void Tin_WithValidData_CreatesSuccessfully()
     {
-        var tin = new Tin("00000000");
+        var tin = Tin.Create("00000000");
         Assert.Equal("00000000", tin.Value);
     }
 
     [Fact]
     public void Tin_WithInvalidLength_ThrowsException()
```
tnickelsen, line 15: Minor: Make Theory and add test for less than 8 digits

tnickelsen (approved):

domains/libraries/EnergyOrigin.Domain/EnergyOrigin.Domain.ValueObjects.Tests/TinTest.cs
martinhenningjensen: Added test test for less than 8 digits, the Theory part, do you mean to collapse all the tests into one With Theory cases?

tnickelsen: Nah just this test case

martinhenningjensen: Done

PR #1611, opened 2024-11-21
explicit treat 00...00 org_id as unauthorized

martinhenningjensen:


domains/libraries/EnergyOrigin.TokenValidation/EnergyOrigin.TokenValidation/B2C/AccessDescriptor.cs
```diff
@@ -14,7 +14,7 @@ public AccessDescriptor(IdentityDescriptor identity)
     public bool IsAuthorizedToOrganization(Guid organizationId)
     {
         var isInternalClient = _identity.SubjectType == SubjectType.Internal;
-        var isOwnOrganization = _identity.OrganizationId == organizationId;
+        var isOwnOrganization = _identity.OrganizationId != Guid.Empty && _identity.OrganizationId == organizationId;
```
TopSwagCode, line 17: Wouldn't it be better to use the orgid we have for 3rd party clients rather than keeping it an empty guid?

TopSwagCode: Furthermore if we get a thirdparty which also is a company which is in denmark. They should be allowed to use API integration to work on their own data?

martinhenningjensen: I agree, but since we don't have danish service providers right now, and I don't know how usable the 3rds party APIs are... I chose to lock-down approach.

martinhenningjensen: I have changed client-credentials orgId claim to contain clients organization id.

TopSwagCode (approved):

PR #1643, opened 2024-12-17
fix bug in measurements parsing

martinhenningjensen:
The fix includes support for 23/24/25 measurements per day for hour based queries and support for 92/96/100 measurements per day for quarter based queries. This will hopefully fix wrong measuremenmts parsing including support for change to/from daylight saving.


domains/measurements/Measurements.API/API/Measurements/gRPC/V1/Services/MeasurementsParser.cs
```diff
@@ -0,0 +1,104 @@
...
+{
+    public List<Measurement> ParseMeasurements(GetMeasurementsRequest request, MeterTimeSeriesResponse dhResponse)
+    {
+        var meteringPoints = dhResponse.GetMeterTimeSeriesResult.MeterTimeSeriesMeteringPoint;
+        var expectedMeasurementCount = (int)(request.DateTo - request.DateFrom) / 3600;
+        var result = new List<Measurement>(Math.Max(expectedMeasurementCount, 0));
+
+        foreach (var meteringPoint in meteringPoints)
+        {
+            // TODO: Assumes a single metering point in each request
```
tnickelsen, line 20: Valid assumption - we query pr gsrn

domains/measurements/Measurements.API/API/Measurements/gRPC/V1/Services/MeasurementsParser.cs
```diff
@@ -0,0 +1,104 @@
...
+            var meteringPointStates = meteringPoint.MeteringPointStates;
+            foreach (var state in meteringPointStates)
+            {
+                // TODO: Assumes a single metering point state for each metering point, this breaks if meter reading occurrence changes in requested period
```
tnickelsen, line 24: what is a metering point state?

domains/measurements/Measurements.API/API/Measurements/gRPC/V1/Services/MeasurementsParser.cs
martinhenningjensen: Remove comment and added filter for correct GSRN.

domains/measurements/Measurements.API/API/Measurements/gRPC/V1/Services/MeasurementsParser.cs
martinhenningjensen: I don't know, but I was told that multiple instances are returned if the meter reading occurrence changes in the requested period.

tnickelsen: MeterReadingOccurance should not change in requested period. Accordring to docs the value can be:
Domain (enumeration):
• PT15M (15 Minutes)
• PT1H (Hourly)
• P1M (Monthly)
• ANDET (Other)
I don't see them change over a request. Do you?

I can't find MeteringPointStates in the docs. We need to figure out what this is I think

martinhenningjensen: The property _MeterReadingOccurance_ contains the actual resolution of the metering point as far as I'm told an able to read in the docs. It's not the resolution requested, hence the bug that I'm trying to fix. 
As far as I'm able to uncover, a separate _metering point state_ is returned for each resolution, so if the metering point changes from ex. hour to quarter resolution in the period we are requesting data from, the multiple states may be returned. I'm haven't seen an example of this, but maybe the excessive logging provides one.

tnickelsen: That's what I mean - how can that happen that during a request it changes from hour to quarter when we specify in the query what we want?

martinhenningjensen: During the period for which data is requested, so if we request data from 1/1 - 10/1 and the metering point changed from hour readings to quarter readings on the 5/1... then as I understand it, we should get multiple metering point states. The docs contains this explanation _'Repeating structure of masterdata changes'_, so if many reasons may give multiple entries I dont know.

tnickelsen: Så hvad gør vi? :)

martinhenningjensen: image

tnickelsen (approved):

PR #1687, opened 2025-02-05
GSRN tostring now correctly prints

tnickelsen:
MeasurementsSyncService line 152: _logger.LogError(e, "An error occured: {error}, no measurements were published, MeteringPoint: {gsrn}", e.Message, syncInfo.Gsrn);

Prints "gsrn":"DataContext.ValueObjects.Gsrn". This should fix that.


martinhenningjensen (approved):
This doesn't really solve the logging issue right?

image

PR #1743, opened 2025-02-24
add inbox/outbox to measurements subsys.

martinhenningjensen:


martinhenningjensen:
https://measurementshotfix.p.acorn-dev.dk/

PR #1778, opened 2025-03-05
fix swapped sender tin/name properties

martinhenningjensen:


domains/transfer/Transfer.API/API/Transfer/Api/[Features]/CreateTransferAgreementCommand.cs
```diff
@@ -73,7 +73,7 @@ public async Task<CreateTransferAgreementCommandResult> Handle(CreateTransferAgr
 
             await UnitOfWork.SaveAsync();
 
-            return new CreateTransferAgreementCommandResult(result.Id, result.SenderTin.Value, result.SenderName.Value,
+            return new CreateTransferAgreementCommandResult(result.Id, result.SenderName.Value, result.SenderTin.Value,
```
Comment (sahma19) [inline domains/transfer/Transfer.API/API/Transfer/Api/[Features]/CreateTransferAgreementCommand.cs:76]: How are they even able to be swapped? is that not what valueobjects are supposed to protect us from?

Sejruppen (approved):

PR #1789, opened 2025-03-12
Preview/4092 get from dh3

tnickelsen:


domains/certificates/Query.API/API/MeasurementsSyncer/Clients/DataHub3/AuthHeaderHandler.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    readonly ITokenService _tokenProvider;
+    readonly AsyncRetryPolicy<HttpResponseMessage> _policy;
+
+    public AuthHeaderHandler(ITokenService tokenProvider)
+    {
+        _tokenProvider = tokenProvider;
+
+        _policy = Policy
+            .HandleResult<HttpResponseMessage>(r => r.StatusCode is HttpStatusCode.Unauthorized or HttpStatusCode.Forbidden)
```
ckr123, line 21: Will this only happen when we get an unauth or forbidden?
Yesterday we talked about how to refresh tokens, why not do it before it get unauth?

domains/certificates/Query.API/API/MeasurementsSyncer/Clients/DataHub3/TokenService.cs
```diff
@@ -0,0 +1,67 @@
...
+    public TokenService(HttpClient httpclient, IOptions<DataHub3Options> dataHub3Options)
+    {
+        _httpClient = httpclient;
+        _dataHub3Options = dataHub3Options;
+        var bodyValues = new Dictionary<string, string>
+        {
+            ["grant_type"] = "client_credentials",
+            ["client_id"] = dataHub3Options.Value.ClientId!,
+            ["client_secret"] = dataHub3Options.Value.ClientSecret!,
+            ["scope"] = "13803f8d-e346-4dae-9738-15fee9efc397/.default"
```
ckr123, line 32: What is this scope?

domains/certificates/Query.API/API/MeasurementsSyncer/Clients/DataHubFacade/DataHubFacadeClient.cs
```diff
@@ -0,0 +1,30 @@
...
+    private readonly HttpClient _client;
+
+    public DataHubFacadeClient(HttpClient client)
+    {
+        _client = client;
+    }
+
+    public async Task<ListMeteringPointForCustomerCaResponse?> ListCustomerRelations(string owner, List<Gsrn> gsrns, CancellationToken cancellationToken)
+    {
+        var mpIdsUrl = string.Join("&meteringPointIds=", gsrns.Select(x => x.Value));
```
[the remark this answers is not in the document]
ckr123: And it looks like we only use it with one MP pr request - will it make sense to only support that?

domains/certificates/Query.API/API/MeasurementsSyncer/RelationHelper.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using API.MeasurementsSyncer.Clients.DataHubFacade;
+using DataContext.ValueObjects;
+
+namespace API.MeasurementsSyncer;
+
+public static class RelationHelper
```
ckr123, line 9: Why create a helper class and not have the validation in the CustomerRelation Class?

domains/certificates/Query.API/API/MeasurementsSyncer/Clients/DataHub3/AuthHeaderHandler.cs
tnickelsen: Hmm that could be a solution as well. This is taken from Heimdal and how they do this - so I'm thinking it's tested. I mean, we will refresh the token on unauthorized and forbidden and the rest of the calls goes through

domains/libraries/EnergyTrackAndTrace.Testing/Extensions/NetworkOptionsExtensions.cs
```diff
@@ -0,0 +1,28 @@
+using System.IO;
+using YamlDotNet.Serialization;
+
+namespace EnergyTrackAndTrace.Testing.Extensions;
+
+public static class NetworkOptionsExtensions
+{
+    public static string ToTempYamlFileUri(this NetworkOptions networkOptions)
```
ckr123, line 8: It doesnt look like this is being used

domains/libraries/EnergyTrackAndTrace.Testing/Extensions/NetworkOptionsExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        var serializer = new SerializerBuilder()
+            .WithNamingConvention(YamlDotNet.Serialization.NamingConventions.CamelCaseNamingConvention.Instance)
+            .Build();
+        var yaml = serializer.Serialize(networkOptions);
+        var path = "file://" + TempFile.WriteAllText(yaml, ".yaml");
+        return path;
+    }
+
+    public static string ToTempYamlFile(this NetworkOptions networkOptions)
```
ckr123, line 18: Why do we need networkoptions to be a yaml file?

domains/libraries/EnergyTrackAndTrace.Testing/PoNetworkOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+    public ChroniclerInfo? Chronicler { get; init; }
+}
+
+public record KeyInfo
+{
+    public required string PublicKey { get; init; }
+}
+
+
+public record ChroniclerInfo
```
ckr123, line 32: Is this needed when we dont use the chrnicler?

ckr123:
Dont we need some test to make sure if there is no relations, metertimes wont be called?

domains/certificates/Query.API/API/MeasurementsSyncer/Clients/DataHubFacade/DataHubFacadeClient.cs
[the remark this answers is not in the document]
tnickelsen: It was integration tested in DataHubFacade: https://github.com/Energinet-DataHub/eo-data-sync/blob/main/services/Datahub.Facade.API/API.IntegrationTests/Services/ListMeteringPointForCustomerCA/ListMeteringPointForCustomerCaTests.cs
Exampled url: /api/relation/meteringpoints/customer?subject={subject}&meteringPointIds=571234567890123456&meteringPointIds=571234567890123457

Hmm I was trying to reflect what is possible with the api instead of restricting it. So in the future, if we need to call with more, we don't need to change the client. We can always call with 1 GSRN

domains/libraries/EnergyTrackAndTrace.Testing/Extensions/NetworkOptionsExtensions.cs
tnickelsen: You are right - this is taken from PO and is used there. So I put it here if needed in the future, while I was at it. Want me to remove it?

domains/libraries/EnergyTrackAndTrace.Testing/Extensions/NetworkOptionsExtensions.cs
tnickelsen: It can be json also, I just chose what we use in PO. It's the global network configuration

domains/libraries/EnergyTrackAndTrace.Testing/PoNetworkOptions.cs
tnickelsen: Nope - removed

domains/libraries/EnergyTrackAndTrace.Testing/TempFile.cs
```diff
@@ -0,0 +1,20 @@
+using System.IO;
+
+namespace EnergyTrackAndTrace.Testing;
+
+public static class TempFile
```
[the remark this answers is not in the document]
tnickelsen: 1 of the functions are used. Again, taken from PO - used for writing the network config to the vault pod. Can delete the other function?

domains/libraries/EnergyTrackAndTrace.Testing/Extensions/NetworkOptionsExtensions.cs
ckr123: I dont think we should have unused code in our solution

domains/libraries/EnergyTrackAndTrace.Testing/TempFile.cs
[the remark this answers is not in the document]
ckr123: The one function is beeing used in a method that is not being used :)

domains/certificates/Query.API/API/MeasurementsSyncer/RelationHelper.cs
ckr123: I still dont see why the helper class is needed, it looks like an anti pattern. Suggestion moving this method to the CustomerRelation class. Doesnt matter if it is in another file. if it is a problem, then i suggest moving customerRelation to a new file

domains/certificates/Query.API/API/MeasurementsSyncer/Clients/DataHub3/TokenService.cs
ckr123: Maybe we should make this question to volt so we dont have magic params in our solution

tnickelsen: Sure, I'll ask

domains/certificates/Query.API/API/MeasurementsSyncer/RelationHelper.cs
tnickelsen: I'm curious - how is it an anti pattern?

Refactored

domains/libraries/EnergyTrackAndTrace.Testing/Extensions/NetworkOptionsExtensions.cs
tnickelsen: removed

domains/libraries/EnergyTrackAndTrace.Testing/TempFile.cs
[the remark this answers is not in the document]
tnickelsen: removed

domains/certificates/Query.API/API/MeasurementsSyncer/Clients/DataHub3/TokenService.cs
tnickelsen: Henrik writes: 'Den fortæller hvilken Azure B2C identity, der er backend application identity for TimeSeriesApi... Den kommer kun til at ændre sig hvis der af en eller anden årsag er brug for at re-deploye denne Identity - eksemplevis i et disaster recovery scenarie'

I put it in the options instead

domains/certificates/Query.API/API/MeasurementsSyncer/RelationHelper.cs
ckr123: The way i see it if a class has knowledge of whether something from another class is valid

ckr123 (approved):

PR #1816, opened 2025-03-24
add cron triggered workflow to update b2c urls

martinhenningjensen:


.github/actions/update-b2c/action.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+        # Build redirect URLs for vcluster branches
+        vclusterBranches=$(echo "${branches}" | grep "vcluster")
+        vclusterDnsPrefixes=$(echo "${vclusterBranches}" | cut -c 9- | tr '[:upper:]' '[:lower:]' | sed 's/[^0-9a-z]*//g')
+        vclusterRedirectUrlsEn=$(echo "${vclusterDnsPrefixes}" | sed "s|^|https://energytrackandtrace-dk.|" | sed "s|$|.v.acorn-dev.dk/en/callback|")
+        vclusterRedirectUrlsDa=$(echo "${vclusterDnsPrefixes}" | sed "s|^|https://energytrackandtrace-dk.|" | sed "s|$|.v.acorn-dev.dk/da/callback|")
+
+        # Add generic redirect URLs
+        redirectUrls=$(echo -e "${previewRedirectUrlsEn}\n${previewRedirectUrlsDa}\n${vclusterRedirectUrlsEn}\n${vclusterRedirectUrlsDa}")
+        redirectUrls=$(echo -e "${redirectUrls}\nhttps://jwt.io")
```
ckr123, line 59: Why is jwt.io and localhost needed?

ckr123 (approved):

.github/actions/update-b2c/action.yaml
martinhenningjensen: For testing, either to inspect the generated tokens or to run the ETT frontend from a developer machine.

.github/actions/update-b2c/action.yaml
```diff
@@ -0,0 +1,72 @@
...
+        previewRedirectUrlsDa=$(echo "${previewDnsPrefixes}" | sed "s|^|https://energytrackandtrace-dk.|" | sed "s|$|.p.acorn-dev.dk/da/callback|")
+
+        # Build redirect URLs for vcluster branches
+        vclusterBranches=$(echo "${branches}" | grep "vcluster")
+        vclusterDnsPrefixes=$(echo "${vclusterBranches}" | cut -c 9- | tr '[:upper:]' '[:lower:]' | sed 's/[^0-9a-z]*//g')
+        vclusterRedirectUrlsEn=$(echo "${vclusterDnsPrefixes}" | sed "s|^|https://energytrackandtrace-dk.|" | sed "s|$|.v.acorn-dev.dk/en/callback|")
+        vclusterRedirectUrlsDa=$(echo "${vclusterDnsPrefixes}" | sed "s|^|https://energytrackandtrace-dk.|" | sed "s|$|.v.acorn-dev.dk/da/callback|")
+
+        # Add generic redirect URLs
+        redirectUrls=$(echo -e "${previewRedirectUrlsEn}\n${previewRedirectUrlsDa}\n${vclusterRedirectUrlsEn}\n${vclusterRedirectUrlsDa}")
```
sahma19, line 58: Will this reveal the actual urls in the job? should we?

.github/workflows/cron-b2c.yaml
```diff
@@ -0,0 +1,22 @@
+name: Update B2C
+
+on:
+  workflow_dispatch: {}
+  schedule:
+    - cron: "*/15 * * * *"
```
sahma19, line 6: Is once a day not fine?

.github/actions/update-b2c/action.yaml
martinhenningjensen: Last lines in the script will dump the updated app configuration:

# Dump app configuration
az ad app show --id ${{ env.ettFrontendAppObjectId }}

.github/actions/update-b2c/action.yaml
```diff
@@ -0,0 +1,72 @@
...
+    required: true
+  azure-client-secret:
+    description: Azure Client secret.
+    required: true
+  github-client-id:
+    description: Github Client ID.
+    required: true
+  github-client-secret:
+    description: Github Client secret.
+    required: true
```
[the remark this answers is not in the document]
martinhenningjensen: App registraion secret will expire just as all our other secrets created in B2C.
Github secrets have been created by Fusion.

.github/workflows/cron-b2c.yaml
martinhenningjensen: Once a day is also possible, its a matter of catching all PRs from ETT and all preview/vcluster environments created by Fusion.

sahma19: This is a CronJob, that executes some logic every 15 minutes. from what i can see it is responsible for deletion?

Somewhere else there is logic that inserts new redirectUrls, into our existing app registration, when a preview/* branch is created. I assume another workflow living in possibly eo-base, or some other place?

So the logic this CronJob executes, move that logic, and place it as a step that happens before inserting new redirect urls, in the other workflow. 

That way, when a developer creates a new branch, the workflow responsible for inserting new redirectUrls, will perform cleanup beforehand.

martinhenningjensen: What would we gain by invoking it from eo-base if the action is placed in energy-origin?

There might be something about access to eo-base repo, but on the other hand B2C client secrets are located in energy-origin where B2C belongs.

I'm not sure what the benefit would be?

sahma19: I am asking since you mentioned this:

But we still need to support environments created in eo-base alone.
---

I understood it as we do not support it?

martinhenningjensen: Ahh, no, it's supported... that is the reason for adding the cron trigger.

PR #1830, opened 2025-03-26
4244 Delete contracts and sliding window on lmc001 errors

tnickelsen:


domains/certificates/Query.API/API.IntegrationTests/Repositories/ContractStateTest.cs
```diff
@@ -180,4 +180,37 @@ public async Task GivenEndedContractWithSlidingWindowWithNoMissingIntervals_When
...
         syncInfo.Should().BeNull();
     }
+
+    [Fact]
+    public async Task DeleteContractAndSlidingWindow_WhenMoreThanOneOfEachOnSameGsrn_DeleteContractAndSlidingWindowMatchingTheGsrn()
+    {
+        // Given ended contract with no missing measurements
+        await using var dbContext = _dbContextFactoryMock.CreateDbContext();
+        var gsrn1 = Any.Gsrn();
+        var gsrn2 = Any.Gsrn();
```
ckr123, line 190: why the gsrn2
Arent you testing two contracts on gsrn1, oth of them are deleted?

domains/certificates/Query.API/API.UnitTests/MeasurementsSyncer/MeasurementsSyncServiceTest.cs
```diff
@@ -20,6 +20,7 @@
 using NSubstitute;
 using Testing.Extensions;
 using Xunit;
+using static System.Runtime.InteropServices.JavaScript.JSType;
```
ckr123, line 23: This can be deleted

domains/certificates/Query.API/API/MeasurementsSyncer/Clients/DataHubFacade/ListMeteringPointForCustomerCaResponse.cs
```diff
@@ -28,4 +28,9 @@ public class Rejection
     public required string ErrorCode { get; set; }
     public required string ErrorDetailName { get; set; }
     public required string ErrorDetailValue { get; set; }
+
+    public bool IsLmc001Error()
+    {
+        return ErrorCode == "LMC-001";
```
ckr123, line 34: Should this have a toLower/upper?

domains/certificates/Query.API/API/MeasurementsSyncer/Persistence/ContractState.cs
```diff
@@ -33,6 +34,19 @@ public async Task<IReadOnlyList<MeteringPointSyncInfo>> GetSyncInfos(Cancellatio
...
     }
 
+    public async Task DeleteContractAndSlidingWindow(Gsrn gsrn)
+    {
+        await using var dbContext = await dbContextFactory.CreateDbContextAsync();
+
+        var contractsToDelete = dbContext.Contracts.Where(x => x.GSRN == gsrn.Value);
+        dbContext.Contracts.RemoveRange(contractsToDelete);
+
+        var slidingWindowsToDelete = dbContext.MeteringPointTimeSeriesSlidingWindows.Where(x => x.GSRN == gsrn.Value);
```
ckr123, line 44: We also have a `SlidingWindowState.cs` should it be two different methods?

domains/certificates/Query.API/API.IntegrationTests/Repositories/ContractStateTest.cs
tnickelsen: Testing that I don't delete more than the 2

domains/certificates/Query.API/API/MeasurementsSyncer/Clients/DataHubFacade/ListMeteringPointForCustomerCaResponse.cs
tnickelsen: Good question. I think it's a good idea, yea. Added

domains/certificates/Query.API/API/MeasurementsSyncer/Persistence/ContractState.cs
tnickelsen: Not sure what you mean by grouping them into one. 
But the thing is, by splitting them into two methods deletion of the two won't happen in the same transaction, unless I pass the same dbcontext as a param to the methods. I can do that but I'm not sure I think it's prettier. What do you think?

tnickelsen: If you look at it, ContractState doesn't handle contracts, it gets sync infos. 
Hmm I could add deletion to existing ICertificateIssuingContractRepository and make a new one for sliding window and add that to the existing unitOfWork and use that. I think making an FK should be because we need it to be an FK and not for a deletion

ckr123 (approved):

PR #1887, opened 2025-04-22
Exclude e2e tests on draft PR

Casper-Olsen:
- E2E tests are running as part of a new workflow
  - They are only running when the PR is in a non draft state. I.e. ready for review

### DIFF
--- a/.github/actions/dotnet-validate-solution/action.yaml
+++ b/.github/actions/dotnet-validate-solution/action.yaml
@@ -17,6 +17,11 @@ inputs:
     default: "false"
     required: false
 
+  test-filter:
+    description: Dotnet test filter
+    default: ""
+    required: false
+
 runs:
   using: composite
 
@@ -46,7 +51,7 @@ runs:
         dotnet tool restore || true
         dotnet restore ${{ inputs.subsystem }}.sln
 
-    - name: Test
+    - name: Test with filter
       shell: bash
       env:
         DOTNET_NOLOGO: "true"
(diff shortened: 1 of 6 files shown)


.github/actions/dotnet-validate-solution/action.yaml
```diff
@@ -48,10 +53,23 @@ runs:
...
+      shell: bash
+      if: ${{ inputs.test-filter != '' }}
+      env:
+        DOTNET_NOLOGO: "true"
+        DOTNET_CLI_TELEMETRY_OPTOUT: "true"
+        DOTNET_SYSTEM_CONSOLE_ALLOW_ANSI_COLOR_REDIRECTION: "true"
+        TERM: "xterm"
+        version: ""
+      run: dotnet test ${{ inputs.subsystem }}.sln --no-restore --configuration Release -warnaserror --logger:"console;verbosity=minimal" --filter ${{ inputs.test-filter }} -p:ParallelizeAssemblies=true -p:ParallelizeTestCollections=true
+
```
sahma19, line 75: Can't this be defined as a single step, with a ternary?

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Print needs
+        run: |
+          echo '${{ toJSON(needs) }}'
+      - name: Verify if merge is allowed - E2E-tests
+        run: |
+          verification=$(echo '${{ toJSON(needs) }}' | jq '.[] | select(.result != "success") | select(.result != "skipped")')
+          if [[ ! -z "$verification" ]]; then
+              echo "Failed"
+              exit 1
+          fi
```
sahma19, line 51: What does this do?

domains/certificates/Query.API/API.IntegrationTests/CertificateIssuingTests.cs
```diff
@@ -37,6 +38,7 @@ public CertificateIssuingTests(IntegrationTestFixture integrationTestFixture) :
     }
 
     [Fact]
+    [E2ETest]
```
sahma19, line 41: Do we only have one e2e?

Casper-Olsen: I just added one test for now, to test that it worked. Not sure which other tests should be added.

.github/workflows/e2e-tests.yml
Casper-Olsen: It runs only E2E tests per sub system, when a pr moves from draft to ready (ready_for_review) and when changes are made to code when the pr is in the ready state.

.github/actions/dotnet-validate-solution/action.yaml
Casper-Olsen: I am doing it in a single step now.

sahma19 (approved):

PR #1904, opened 2025-05-09
Preview/org disable in auth

sahma19:


domains/authorization/Authorization.API/API/Authorization/EventHandlers/AuthorizationOrganizationRemovedFromWhitelistEventHandler.cs
```diff
@@ -22,6 +23,10 @@ public async Task Consume(ConsumeContext<OrganizationRemovedFromWhitelist> conte
         logger.LogInformation("Organization {orgId} removed from whitelist, removing all clients", orgId);
         var removeOrganizationClientsCommand = new RemoveOrganizationClientsCommand(orgId);
         await mediator.Send(removeOrganizationClientsCommand, context.CancellationToken);
+
+        logger.LogInformation("Organization {orgId} removed from whitelist, removing all clients", orgId);
```
Casper-Olsen, line 27: I think this should be a different log message?

domains/authorization/Authorization.API/API/Authorization/Exceptions/WalletNotEnabled.cs
```diff
@@ -0,0 +1,5 @@
+using System;
+
+namespace API.Authorization.Exceptions;
+
+public class WalletNotEnabled(string message) : Exception(message);
```
Casper-Olsen, line 5: I think this will lead to an internal server error.
'ExceptionHandlerMiddleware' needs to handle the exception.
It might be better to just use a 'BusinessException' if you don't need the specific exception type.

Casper-Olsen (changes requested):

domains/authorization/Authorization.API/API/Authorization/EventHandlers/AuthorizationOrganizationRemovedFromWhitelistEventHandler.cs
sahma19: oops yes :D

sahma19: Fixed: https://github.com/Energinet-DataHub/energy-origin/pull/1904/commits/fb79f2f17ed0eec12392a2cd7f7cf384abbb4a77

domains/authorization/Authorization.API/API/Authorization/Exceptions/WalletNotEnabled.cs
sahma19: Fixed: https://github.com/Energinet-DataHub/energy-origin/pull/1904/commits/2137b734964f9d8670a27c2a6ddef5800c3ded50

Casper-Olsen (approved):

PR #2087, opened 2025-06-26
4667 OperationCancelled

ckr123:


domains/transfer/ClaimAutomation/Worker/Automation/ClaimWorker.cs
```diff
@@ -32,7 +34,43 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+            catch (OperationCanceledException)
+            {
+                logger.LogWarning("ClaimWorker has been cancelled");
+            }
+            catch (Exception e)
+            {
+                switch (e)
+                {
+                    case HttpRequestException { StatusCode: HttpStatusCode.BadRequest }:
+                        logger.LogWarning("Wallet returned bad request: {exception}", e);
```
Casper-Olsen, line 53: Are we calling wallet here?
I can see the log error refers to ClaimService.

ckr123: Yes we are calling wallet to claim and to get certificates.
I cant remember why this was implemented. This change is only due to a copy/paste

Casper-Olsen (approved):

PR #2130, opened 2025-07-10
Make sure that contracts use the correct MeteringPointType

Casper-Olsen:


domains/measurements/Measurements.API/API/MeteringPoints/Api/Dto/Responses/MeteringPoint.cs
```diff
@@ -9,7 +9,7 @@ namespace API.MeteringPoints.Api.Dto.Responses;
 public record MeteringPoint(
     string GSRN,
     string GridArea,
-    MeterType MeteringPointType,
+    MeterType Type,
```
sahma19, line 12: Is this the "best" solution?

sahma19 (dismissed):

PR #2171, opened 2025-08-07
Only allow trial users to create and edit contracts through the contracts controller

Casper-Olsen:
# Main feature:
- Only allow trial users to create and edit contracts through the contracts controller

# Refactor:
-  Use same models for both MeteringPoint clients

# Documentation
- Added contract docs again, with disclaimer about trial (it can only be used from trial)
- Added to create and edit contract endpoints description, that they can only be used by trial organizations
- _Screenshots of documentation at the end of PR description_

# Test:
- E2E test create contract through internal endpoint

# Misc:
- Removed .gitattributes since it does not solve the line ending issue for documentation.xml files and I was having issues with new xml docs not being generated

# Documentation screenshots:
image

image


domains/certificates/Query.API/API/Query.API/Controllers/ContractsController.cs
```diff
@@ -44,6 +44,11 @@ public async Task<ActionResult> CreateContract(
             return Forbid();
         }
 
+        if (!identityDescriptor.IsTrial())
```
ckr123, line 47: Could you add a test to verify this?

Casper-Olsen: There is a test for it:
CreateContract_WithLiveOrganization_Forbidden

domains/certificates/Query.API/API/Query.API/Controllers/ContractsController.cs
```diff
@@ -138,6 +143,11 @@ public async Task<ActionResult> UpdateEndDate(
...
+        {
```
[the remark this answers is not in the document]
Casper-Olsen: There is a test for it:
EditEndDate_WithLiveOrganization_Forbidden

